PR #7, opened 2014-10-30
Making base64 encoding/decoding optional

arminc:
I am storing my messages as Json in to the database and I didn't want them to be base64 so I made it optional.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ organization := "com.github.dnvriend"
 
 name := "akka-persistence-jdbc"
 
-version := "1.0.8"
+version := "1.0.9-SNAPSHOT"
 
 scalaVersion := "2.11.2"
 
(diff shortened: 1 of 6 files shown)


prettynatty:
It also would be great to have ability to store messages in column with BLOB type (bytea in postgres for example).

t3hnar:
hi, @dnvriend are you going to merger this?

dnvriend:
This week I will look at all the pull requests. The serialization feature is something I want to have for some time now, but I am thinking about a pluggable architecture, so this pull request will not be it but I will look into it. I am glad that you contributed though and your code will serve for inspiration. Thank you! 

Regards

Dennis

t3hnar:
Thanks, just wanted to add some feature that are nice to have:
* ability to change `journal`/`snapshot` tables schema and mapping
* it should be easy to store data in readable formats (json, xml, etc..)

Really looking forward to your new release :)

Issue #9, opened 2014-12-05
java.sql.SQLRecoverableException: No more data to read from socket


stevenrigbyS4N:
Hello,

The project I'm currently working on uses akka cluster sharding with this plugin connected to an Oracle database. It works fine but after leaving the project running for some time or when the database is stopped it throws the following error and it is unable to recover:

2014/12/05 09:25:23.887 ERROR[t-dispatcher-14] s.StatementExecutor$$anon$1 81 --- SQL execution failed (Reason: No more data to read from socket):

   MERGE INTO USERSIS.snapshot snapshot USING (SELECT '/user/sharding/RequestIDRouterCoordinator/singleton/coordinator' AS persistence_id, 7 AS seq_nr from DUAL) val ON (snapshot.persistence_id = val.persistence_id and snapshot.sequence_nr = val.seq_nr) WHEN MATCHED THEN UPDATE SET snapshot='qAAAAKztAAVzcgAtYWtrYS5wZXJzaXN0ZW5jZS5zZXJpYWxpemF0aW9uLlNuYXBzaG90SGVhZGVyAAAAAAAAAAECAAJJAAxzZXJp... (1268)' WHEN NOT MATCHED THEN INSERT (PERSISTENCE_ID, SEQUENCE_NR, SNAPSHOT, CREATED) VALUES ('/user/sharding/RequestIDRouterCoordinator/singleton/coordinator', 7, 'qAAAAKztAAVzcgAtYWtrYS5wZXJzaXN0ZW5jZS5zZXJpYWxpemF0aW9uLlNuYXBzaG90SGVhZGVyAAAAAAAAAAECAAJJAAxzZXJp... (1268)', 1417789523876)

2014/12/05 09:25:23.888 ERROR[n-dispatcher-44] s.StatementExecutor$$anon$1 81 --- SQL execution failed (Reason: No more data to read from socket):

   DELETE FROM USERSIS.snapshot WHERE persistence_id = '/user/sharding/RequestIDRouterCoordinator/singleton/coordinator' AND sequence_nr = 7

[ERROR] [12/05/2014 09:25:23.889] [TokenCluster-akka.actor.default-dispatcher-14] [akka://TokenCluster/system/snapshot-store] No more data to read from socket
java.sql.SQLRecoverableException: No more data to read from socket
	at oracle.jdbc.driver.T4CMAREngine.unmarshalUB1(T4CMAREngine.java:1157)
[...]

PR #10, opened 2014-12-27
Adds ´validationQuery´ configuration parameter

miguel-vila:
This adds a new `validationQuery` configuration parameter. If the `jdbc-connection` configuration doesn't have a `validationQuery` field then it is set to `null` (`ConnectionPoolSettings` would still work as expected [according to its default value](https://github.com/scalikejdbc/scalikejdbc/blob/develop/scalikejdbc-core/src/main/scala/scalikejdbc/ConnectionPoolSettings.scala#L25)).

This may solve issue #9 according to [a suggestion  by Victor Klang](https://groups.google.com/d/msg/akka-user/xRappYKjoho/3sYnCk0X9K8J). We are having this issue when using Oracle, so (for us) the solution maybe is to have a `validationQuery`config field set to `select 1 from dual`. I'm not sure if this is the solution, I'm only following Victor's suggestion. Also, I'm not sure if this same problem occurs with other DBs.

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/common/ActorConfig.scala
+++ b/src/main/scala/akka/persistence/jdbc/common/ActorConfig.scala
@@ -27,6 +27,7 @@ class PluginConfig(system: ActorSystem) {
   def journalTableName = config.getString("journalTableName")
   def snapshotSchemaName: String = config.getString("snapshotSchemaName").toOption.map(_ + ".").getOrElse("")
   def snapshotTableName = config.getString("snapshotTableName")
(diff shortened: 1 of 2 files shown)


prettynatty:
@dnvriend are you going to merge this PR?

prettynatty:
It also would be great to have ability to store messages in column with BLOB type (bytea in postgres for example).

dnvriend:
I will merge it this week, stay tuned!

bpg:
It seems that this fix did not help in my case, see my [comment](https://github.com/dnvriend/akka-persistence-jdbc/issues/9#issuecomment-70948112) to #9

PR #20, opened 2015-05-06
use apache commons codec Base64 vs the java8-only java.util.Base64

mwkohout:
java.util.Base64 was only created in Java 8.  By using apache commons, this library can still be built against Java 7.


dnvriend:
Merged, thanks for your contribution!

PR #21, opened 2015-05-07
use a ParameterBinder to pass snapshot into the merge statement and get rid of the stored procedure.

mwkohout:
Some DBAs frown on stored procedures.  This change eliminates the need for the stored procedure.  

And locally some tests were failing for me because timeouts were too short.  I increased some of the timeouts in those tests.

I haven't updated the library's documents, so let me know if you want that as part of the change

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/snapshot/Statements.scala
+++ b/src/main/scala/akka/persistence/jdbc/snapshot/Statements.scala
@@ -1,5 +1,8 @@
 package akka.persistence.jdbc.snapshot
 
+import java.io.{StringReader, ByteArrayInputStream, StringWriter, InputStream}
+import java.sql.PreparedStatement
+
 import akka.persistence.jdbc.common.PluginConfig
 import akka.persistence.jdbc.serialization.{SnapshotSerializer, SnapshotTypeConverter}
 import akka.persistence.serialization.Snapshot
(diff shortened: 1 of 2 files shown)


bpg:
Oracle's MERGE won't work for BLOBs greater than some internal buffer size. See #9 for more details.

mwkohout:
Have you executed the tests in your environment?  In mine they all pass including 'Macbeth'.

If not, what version of the oracle drivers are you using?

bpg:
It was 100% reproducible in the project's test environment on v 1.0.9. You can checkout bfa643bf334957c4dc01895839ba5eabb8f0be2e and run `test-oracle.sh`.

mwkohout:
I don't disagree that passing the clob as a string was incorrect, but this patch works differently.

Have you checked out my alterations and run them?  Also, if you have and they failed, could you share what version of Oracle you ran them against?

mwkohout:
I noticed that these tests weren't actually asserting the value of the saved snapshots when they were retrieved.

dnvriend:
Thanks for your contribution and the discussion. Is there any problem with the code, the way it is, and/or is there any risk changing it back to the `MERGE INTO` statement? If not, I will merge the pull request.

bpg:
The code from your branch seems to be working in all the tests. I think the `ParameterBinder` did the trick. 
Thanks for fixing that, I don't really like the SP approach as well :)

mwkohout:
I don't think there are any new risks.  But I think those same risks would exist when using the stored procedure version of this code..

dnvriend:
Thank you both. That settles it I will merge it asap.

PR #27, opened 2015-10-11
don't fail on asyncWrite with empty messages

prettynatty:
While debugging weird bugs in my app I traced them to akka-persistence-jdbc:
I noticed that it generates like this:
```
INSERT INTO akka.journal (persistence_id, sequence_number, marker, message, created) VALUES 
```
Scalikejdbc logs error and after that application falls into the ask-timeout mess. I went through akka sources and if I understood right, it may execute asyncWriteMessages with empty messages list. So, here is the fix.

Please, release new version including it.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,7 +3,7 @@ organization := "com.github.dnvriend"
 
 name := "akka-persistence-jdbc"
 
-version := "1.2.0"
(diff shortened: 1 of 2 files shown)


dnvriend:
Thanks for reporting and for this workaround!

PR #37, opened 2016-03-17
Make offset sequential on eventsByTag queries

### LINKED ISSUE #36: Offset in ReadJournal eventsByTag query is actually sequenceNr
The README mentions:
> You can retrieve a subset of all events by specifying offset, or use 0L to retrieve all events with a given tag. **The offset corresponds to an ordered sequence number for the specific tag.** Note that the corresponding offset of each event is provided in the EventEnvelope, which makes it possible to resume the stream at a later point from a given offset.

However what is actually used for the offset field of the EventEnvelope is the persistenceId's sequenceNr as it can be seen in the [ReadJournal implementation](https://github.com/dnvriend/akka-persistence-jdbc/blob/master/src%2Fmain%2Fscala%2Fakka%2Fpersistence%2Fjdbc%2Fquery%2Fjournal%2Fscaladsl%2FJdbcReadJournal.scala#L77):
```scala
  override def currentEventsByTag(tag: String, offset: Long): Source[EventEnvelope, NotUsed] =
    journalDao.eventsByTag(tag, offset)
      .via(serializationFacade.deserializeRepr)
      .mapAsync(1)(deserializedRepr ⇒ Future.fromTry(deserializedRepr))
      .map(repr ⇒ EventEnvelope(repr.sequenceNr, repr.persistenceId, repr.sequenceNr, repr.payload))
```

Looks like either the documentation or the implementation is wrong, or I may be missing something quite terribly here.

wwwiiilll:
Resolves #36 

The implementation assumes that the `offset` passed to the `eventsByTag` query will be <= to the actual event count since it'll be the starting point for the future events no matter how many were actually skipped.
This is enough for my usage and maybe as far as I can go with my very limited Akka Streams knowledge, but something better could be designed.

I restored the note about ordering in the README since it is still applicable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Configure `slick`:
   - `slick.driver.PostgresDriver`
   - `slick.driver.MySQLDriver`
   - `com.typesafe.slick.driver.oracle.OracleDriver`
-   
+
 ## DataSource lookup by JNDI name
 The plugin uses `slick` as the database access library. Slick [supports jndi](http://slick.typesafe.com/doc/3.1.1/database.html#using-a-jndi-name)
 for looking up [DataSource](http://docs.oracle.com/javase/7/docs/api/javax/sql/DataSource.html)s. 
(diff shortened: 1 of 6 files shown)


dnvriend:
Thanks for your contribution!

PR #46, opened 2016-04-25
Significant performance boost by using compiled queries

### LINKED ISSUE #45: Use compiled queries
Slick query compiler is very slow in comparsion with its compiled queries. Compiled queries usage gives >10x performance boost.
I'll do a pull request.

prettynatty:
fixes #45

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,7 +18,7 @@ name := "akka-persistence-jdbc"
 
 organization := "com.github.dnvriend"
 
-version := "2.2.18"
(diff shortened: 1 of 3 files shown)


dnvriend:
Thanks, nice catch, without the fix the performance test would fail, and with the fix it passes. I won't release it just yet, there seems to be a problem with the query API that I want to fix first.

dnvriend:
Thanks! I have created a v2.2.19 release with the fixes!

PR #47, opened 2016-04-26
Ability to get Database instance from JNDI

prettynatty:


prettynatty:
How about this one?
It's nessesary to have an ability to share AsyncExecutor and other Database settings with plugin.

dnvriend:
Thanks! I have created a v2.2.19 release with the fixes

PR #54, opened 2016-06-07
Honour the schemaName setting for the snapshot table

charithe:
The `schemaName` setting in the configuration is currently being ignored for the snapshot table. This causes problems when using a schema other than `public` with Postgres (`ERROR: relation "snapshot" does not exist`).


dnvriend:
Thanks for noticing and your contribution! I will merge and release ASAP

PR #56, opened 2016-06-18
Fcristovao/revamp serialization

fcristovao:
Hey there,

I've removed the whole of the previous serialization mechanism and introduced 2 new classes, `PersistentReprSerializer`and `SnapshotSerializer`. Then's it's up to the DAOs to implement those how it's best for them.

I also removed the *serialization* option from the plugin. From now on, serialization is on, as long as the DAO so decides to use it (which is the case now).

I had to pass in the `Serialization` to the DAOs, since, ultimately, it's akka who does the serialization, but the DAOs have specific serializers (that they themselves control). A DAO can decide not to use what's passed on to it.

I also moved each "module" of the ByteArray DAO to it's own package. This should aid in a future split.
I did some other minor fixes (in particular I introduced a "beforeAll" clause in the `JournalSpec` and `JournalPerfSpec`, since I was having a Heisenbug whenever I ran the Postgres ones together).

All tests are green now :)


dnvriend:
Thanks, it looks really clean!

PR #57, opened 2016-06-19
Fix for usage with h2

fcristovao:
Introduces the capability to use H2 as a DB.

It adds all the tests for it (and I revamped them a bit to avoid duplication).

H2 is particularly useful for component testing without the need of having a docker container already running :)

Please check my last commit. It was kind of an experiment (using stackable traits). In my view it makes things a bit cleaner (with separation of concerns for each driver), and technically it's even (a tiny bit) more performant (negligible).
I found it interesting how in the base traits it showed that not as much inputs were needed :)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -75,9 +75,8 @@ scalacOptions ++= Seq(
 )
 
 // enable scala code formatting //
-import com.typesafe.sbt.SbtScalariform
-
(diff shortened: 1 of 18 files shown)


dnvriend:
Thanks for your contribution. This is going to help users that want to use H2 a lot!

PR #58, opened 2016-07-05
Fixed invalid journal table name when schemaName config is empty

jtysper:
In case the `schemaName` config is empty, `#${schemaName.getOrElse("")}"."#$tableName` will result in `.tableName` instead of just `tableName`.
Added a `fullTableName` value to config that omits the `.` in case of empty `schemaName`.

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/config/AkkaPersistenceConfig.scala
+++ b/src/main/scala/akka/persistence/jdbc/config/AkkaPersistenceConfig.scala
@@ -44,6 +44,7 @@ class JournalTableConfiguration(config: Config) {
   private val cfg = config.asConfig("tables.journal")
   val tableName: String = cfg.as[String]("tableName", "journal")
   val schemaName: Option[String] = cfg.as[String]("schemaName").trim
+  val fullTableName = schemaName.map(_ + ".").getOrElse("") + tableName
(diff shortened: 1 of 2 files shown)


jtysper:
Am I doing something wrong or are the unit-tests currently not working?

dnvriend:
Yes I know, after adding the ordering column, I get problems with Oracle with the auto-increment column as Oracle does not support it. I you know how to fix it I would be very grateful! What does work is Postgres, MySQL and H2.

jtysper:
To be honest, I don't really get the idea behind the `ordering` column.
Isn't it kind of a duplication of the `created` column?

I figured some tests are failing because the `offset` value of `EventEnvelope` is initialized with the `ordering`. The thing with the current `offset` value is, that it's a strictly growing number over **all** events, so it's not really representing the offset within a tag-query.

dnvriend:
The created column is not usable to get the total ordering of events in the journal as the created value can be the same when the journal plugin does a bulk-insert. For the *byTag query we need the total ordering, so a new column is added

jtysper:
Ah, ok. Makes sense.

jtysper:
Hi, sorry for the delay, I was on vacation.
What about the schemaName fix for empty schema config?

dnvriend:
Hi, I hope you had a very nice vacation. Its on my backlog. I will check if its still an issue.

PR #61, opened 2016-07-27
Sql optimizations

rockjam:
I recently noticed recovery time degradation with growth of journal table. One of unnecessarily slow queries I found was selection of last sequence number for both journal and snapshot.

```
selectAllJournalForPersistenceId(persistenceId).map(_.sequenceNumber).max 
```
translates into this sql:

```
select max(jour.sequence_number) from (select persistence_id, sequence_number from journal where persistence_id = 'Neighbour-10' order by sequence_number desc) jour;
```

which will become slower with growth of persisted events for given persistence_id(1+ seconds on journal with total 5 000 000 records), whereas all we need is to get last sequence_number. Simpler equivalent request is:

```
selectAllJournalForPersistenceId(persistenceId).map(_.sequenceNumber).take(1) 
```

translates into this sql:
```
select sequence_number from journal where persistence_id = 'Neighbour-10' order by sequence_number desc limit 1;
```

Same changes applied to snapshot queries.

Also, because there are a lot of queries filtered by `persistence_id` and `sequence_number` in journal table index on (persistence_id, sequence_number) required.

### DIFF
--- a/src/main/resources/schema/h2/h2-schema.sql
+++ b/src/main/resources/schema/h2/h2-schema.sql
@@ -10,6 +10,8 @@ CREATE TABLE IF NOT EXISTS PUBLIC.journal (
   PRIMARY KEY(ordering, persistence_id, sequence_number)
 );
 
+CREATE INDEX journal_persistence_id_sequence_number_idx ON PUBLIC.journal(persistence_id, sequence_number);
+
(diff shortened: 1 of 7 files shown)


dnvriend:
Thanks for your PR! Great job on optimizing the queries. I will release it shortly.

PR #63, opened 2016-07-29
Oracle 12 c

jtysper:
Okay, finally... :-)

First of all: **Thanks for this incredible amount of effort you put into this community plugin!**
I learned so much doing these changes and really appreciate your work!

A word of warning: The MySQL Tests didn't quite work after I rebased to 2.6.3, I was always running into timeouts within the Actor tests...

Also please have an **extra extra** careful look at this PR, since I am not really sure I knew exactly what I was doing all the time and relied mostly on the existing tests... so be warned! ;-)

To summarize what I did:

- changed the bash scripts to not just delete all not running docker containers but instead use docker-compose - for people running docker natively that might have been a big surprise in the past
- added host-config for cases when not using boot2docker
- went back to Oracle 11g, since I ran into problems with the 12c docker image (took years to start)
- switched the eventsByTags logic to use `ordering` as `offset` instead of `rownum`
- changed logic of fullTableName so that the `.` doesn't appear in front of the tableName when there's no schemaName
- added bintraySbt plugin because sbt wouldn't build on my machine
- renamed the sequence index, because Oracle 11g only allows 35 chars in names (...)
- fixedup Oracle db-cleanup for unit-tests (e.g. it didn't reset the sequence so the offset was wrong for tests down the line)
- removed @ignore from OracleJournalPerfSpec

Things i'd **really** like you to check before merging anything:

- There have been quite some changes regarding `deleted` and `created` etc. Not sure if the Oracle schema/code is actually up-to-date now - ~~e.g. there's still the deleted_to table and I didn't remove it~~ sorry, just realized actually I did
- I am not sure how good the test-coverage actually is, so I am not sure if I broke corner cases with my changes so you might be the way better judge on this
- If I did anything stupid, please tell me - I'm really new to this field (Scala, FP, Typesafe Reactive Platform etc.) so the more I can learn from way more experienced guys like you, the better for me :-)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 rm ./bintray.sbt
-sbt "test-only *Postgres* *MySQL* *H2*"
+sbt test
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -25,3 +25,5 @@ addSbtPlugin("de.heikoseeberger" % "sbt-header" % "1.5.1")
 // generates Scala source from your build definitions //
 // see: https://github.com/sbt/sbt-buildinfo
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.6.1")
+
+addSbtPlugin("me.lessis" % "bintray-sbt" % "0.3.0")
--- a/scripts/docker-compose.yml
+++ b/scripts/docker-compose.yml
@@ -1,6 +1,6 @@
 postgres:
   image: postgres:latest
-  container_name: postgres
+  container_name: postgres-test
   environment:
     - "TZ=Europe/Amsterdam"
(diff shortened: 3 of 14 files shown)


jtysper:
Ok, I just realized that the Oracle test-cases haven't been executed on travis.
I added the Oracle to build.sh (which I completely ignored previously) but the tests aren't able to connect to the db. Is this because you run the DB servers via travis and not via docker?

dnvriend:
Hi, you have been busy! It would be great to have oracle support back in the plugin and thanks to you this may actually be the case, thank you very much! I will take a look at the PR and let you know if I find anything. Stay tuned!

dnvriend:
Looks good to me!

PR #66, opened 2016-08-20
eventsByPersistenceId should terminate when toSequenceNr is reached.

monktastic:
I could be wrong about the spec, but I noticed that eventsByPersistenceId(m, n) does not complete even when n <= the max journal sequence number. Other impls (including leveldb) do terminate.

What do you think?


dnvriend:
Hi, thanks for the PR, I will look into the issue.

monktastic:
Thanks Dennis. I've started a discussion here: https://groups.google.com/forum/#!topic/akka-user/hl2A6e65gsQ

dnvriend:
Hi, I have read the discussion and have looked into the issue and tested it on the leveldb implementation of akka v2.4.9. The LevelDb version does not complete when toSeqNr==0, it however should. Your fix is quite elegant. I will merge the PR. Thanks!

monktastic:
Ah yes, we ran into that bug as well.

Thanks for the quick response and merge!

PR #72, opened 2016-09-07
Snapshot table should only define one primary key

### LINKED ISSUE #71: Snapshots table defines three primary keys
The snapshots table defines three primary keys, which is illegal:

``` scala
  class Snapshot(_tableTag: Tag) extends Table[SnapshotRow](_tableTag, _schemaName = snapshotTableCfg.schemaName, _tableName = snapshotTableCfg.tableName) {
    def * = (persistenceId, sequenceNumber, created, snapshot) <> (SnapshotRow.tupled, SnapshotRow.unapply)

    val persistenceId: Rep[String] = column[String](snapshotTableCfg.columnNames.persistenceId, O.Length(255, varying = true), O.PrimaryKey)
    val sequenceNumber: Rep[Long] = column[Long](snapshotTableCfg.columnNames.sequenceNumber, O.PrimaryKey)
    val created: Rep[Long] = column[Long](snapshotTableCfg.columnNames.created)
    val snapshot: Rep[Array[Byte]] = column[Array[Byte]](snapshotTableCfg.columnNames.snapshot)
    val pk = primaryKey("snapshot_pk", (persistenceId, sequenceNumber))
  }
```

Essentially this is saying that the persistenceId column is the primary key, the sequenceNumber column is the primary key, and there is also a composite primary key of the two of them.  Here's an example schema that Slick will generate for this:

``` sql
create table "snapshot" (
  "persistence_id" VARCHAR(255) NOT NULL PRIMARY KEY,
  "sequence_number" BIGINT NOT NULL PRIMARY KEY,
  "created" BIGINT NOT NULL,
  "snapshot" BLOB NOT NULL
);
alter table "snapshot" add constraint "snapshot_pk" primary key("persistence_id","sequence_number");
```

No database will execute this successfully since there are three primary keys defined.  The `O.PrimaryKey` option should be removed from the column definitions.

jroper:
Fixes #71.

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/snapshot/dao/SnapshotTables.scala
+++ b/src/main/scala/akka/persistence/jdbc/snapshot/dao/SnapshotTables.scala
@@ -39,8 +39,8 @@ trait SnapshotTables {
   class Snapshot(_tableTag: Tag) extends Table[SnapshotRow](_tableTag, _schemaName = snapshotTableCfg.schemaName, _tableName = snapshotTableCfg.tableName) {
     def * = (persistenceId, sequenceNumber, created, snapshot) <> (SnapshotRow.tupled, SnapshotRow.unapply)
 
-    val persistenceId: Rep[String] = column[String](snapshotTableCfg.columnNames.persistenceId, O.Length(255, varying = true), O.PrimaryKey)
-    val sequenceNumber: Rep[Long] = column[Long](snapshotTableCfg.columnNames.sequenceNumber, O.PrimaryKey)
+    val persistenceId: Rep[String] = column[String](snapshotTableCfg.columnNames.persistenceId, O.Length(255, varying = true))
+    val sequenceNumber: Rep[Long] = column[Long](snapshotTableCfg.columnNames.sequenceNumber)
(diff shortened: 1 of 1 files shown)


dnvriend:
This PR breaks the UPSERT functionality of slick, please check.

See my anwer on issue #71 

_Hi, this is true, but for the fact that this workaround is necessary to get the UPSERT functionality working with Slick see [Slick Issue #966 - Scala's Slick with multiple PK insertOrUpdate() throws exception ERROR: syntax error at end of input](https://github.com/slick/slick/issues/966), as it stands, we must wait for slick 3.2.0_

jroper:
No worries, we can close it for now if you want.

dnvriend:
Agreed.

daniel-moss-1:
Hi @dnvriend - now that the plugin makes use of slick 3.2.0, could this PR be looked at again?

dnvriend:
Put on the backlog

rcavalcanti:
@jroper this PR is probably ok now because the master branch is using Slick 3.2.1. 
GitHub is showing conflict, but I could merge locally without problem. And all tests passed. 

Could you have a quick look into it?

jroper:
Rebased.

PR #75, opened 2016-09-07
Removed binary dependency on slick-extensions

jroper:
slick-extensions 3.1.0 is a proprietary library (yes, Lightbend has open sourced it, but that doesn't retroactively change the license for existing releases). Technically, any project that includes it on its classpath in production that doesn't have a Lightbend subscription is in breach of the Lightbend license.

This PR solves this problem by removing the binary dependency on the Oracle driver, allowing akka-persistence-jdbc to be used in production without being in violation of Lightbend's license.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,13 +12,16 @@ Akka-persistence-jdbc writes journal and snapshot entries entries to a configure
 Add the following to your `build.sbt`:
 
 ```scala
+libraryDependencies += "com.github.dnvriend" %% "akka-persistence-jdbc" % "2.6.6"
+```
+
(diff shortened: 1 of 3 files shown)


dnvriend:
Thanks, I have interpreted the library being changed to Open Sourced wrongly. I will change the readme accordingly...

skisel:
Is it correct that as soon as slick 3.2 is released, and akka-persistence-jdbc will support it, the problem with Oracle drivers will be gone because they will be included as part of slick?

jroper:
That's correct.

PR #78, opened 2016-10-09
Snapshot storage BLOB handling.

skisel:
Here is PR to fix the problem with inserting blobs more than 4K in size. More details here:

https://github.com/dnvriend/akka-persistence-jdbc/issues/77

Please apologize if my code does not match your style, I will be happy to hear your feedback and fix it.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -37,6 +37,7 @@ libraryDependencies ++= {
     "com.github.dnvriend" %% "akka-persistence-query-writer" % "0.0.2",
     "com.typesafe.akka" %% "akka-stream" % akkaVersion,
     "com.typesafe.slick" %% "slick" % slickVersion,
+    "commons-io" % "commons-io" % "2.4",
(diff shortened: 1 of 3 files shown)


skisel:
Ok I see, I broke the test case which expected to override snapshot. Will submit the fix soon.

skisel:
I've submitted another change to the PR. I fixed the broken test and made Oracle to be working with big blobs (more than 4k bytes).

dnvriend:
Thanks for the PR, I will put it on the bug list for now.

skisel:
I think the test case is not needed anymore, cause I was able to push for the test case in akka testkit:

https://github.com/akka/akka/issues/21704#event-838170362

It has been released as part of akka 2.4.12 release. I will update my PR.

dnvriend:
Thanks for all your effort. This fix will be included in the next release!

PR #85, opened 2016-11-27
Migrated to Slick 2.3.0-M1

joseblas:
Migration to Slick 3.2.0-M1 (No need of Slick Extensions and its expensive subscription)


dnvriend:
Thanks, I have taken your code and released a test version `2.6.12-3.2.0-M2`.

PR #88, opened 2016-12-23
compiled inserts

prettynatty:


dnvriend:
Its a bit late. I was kinda busy with other things, so the akka-persistence-jdbc had low priority. As always, thanks for your contribution!

Issue #65, opened 2016-08-08
Event adapters not applied on `eventsByTag`


jvmakine:
Hi! I'm migrating our database from Cassandra to Postgres, and with that, our connector to `akka-persistence-jdbc`. The problem I am having is that it seems that, unlike with the Cassandra connector, the event adapters are not being called when reading tagged events from `JdbcReadJournal`. My configuration is as follows (omitting the slick configuration):

```
jdbc-journal {   
  slick = ${slick} 
  recovery-event-timeout = 60m  
  event-adapters {   
    adapter = "com.x.Adapter"   
  }    
  event-adapter-bindings {     
    "com.x.PersistedEvent" = adapter   
  }
 }  

jdbc-snapshot-store {   
  slick = ${slick} 
} 

 jdbc-read-journal {   
  refresh-interval = "100ms"   
  max-buffer-size = "500"   
  batch-size = "250"   
  slick = ${slick} 
}
```

The adapter is being called correctly when writing events, but not when reading them from `JdbcReadJournal`.

Am I misunderstanding something, missing something from my configuration, or are the event adapters not supported when reading from the `JdbcReadJournal` with `eventsByTag`?

I am using version `2.6.5-RC2`

dnvriend:
Hi, your configuration looks correct. AFAIK, event adapters are only applicable for persistence actors, they are for mapping the application domain messages on the data domain messages and vice versa. When querying the journal directly, you get an event envelope and you'll have to handle the data model directly. I have tried to create an image how it works: 

![event-adapters](https://github.com/dnvriend/akka-serialization-test/raw/master/img/akka-persistence-event-adapters.png)

jvmakine:
Thank you very much for the fast response and clarifying this. I'll refactor our code accordingly and close this issue.

filosganga:
@dnvriend The fact that you receive an EventEnvelope does not interfere with the event adaptation. If the query does not adapt events, it is completely useless as many migration logics are inside the EventAdapters.

FYI, The  Cassandra plugin and LevelDb plugin, that re the official ones, run the adapters also during the query.

But of course the decision is up to you.

dnvriend:
Hi, thanks for the heads-up. I'll take a look.

cwei-bgl:
@dnvriend I wish to have event adapters working with persistence query as well. It looks ugly to have these adapters in ReadJournal process. Thanks.

Alnaimi-:
[...]

PR #103, opened 2017-05-11
Added support for event adapters on query side.

aenevala:
Refs #65


src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -56,6 +59,8 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem)
   val db = SlickDatabase.forConfig(config, readJournalConfig.slickConfiguration)
   sys.addShutdownHook(db.close())
 
+  val eventAdapters = Persistence(system).adaptersFor(JdbcAsyncWriteJournal.Identifier)
```
aenevala: Other option is to fetch the identifier from the configuration.

dnvriend:
Hi, thanks for the PR, I'll take a look at it

src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
```diff
@@ -33,6 +33,9 @@ import scala.collection.immutable._
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
+object JdbcAsyncWriteJournal {
+  val Identifier = "jdbc-journal"
```
dnvriend: final val provides byte code optimization for these kinds of identifiers.  Try it out in the REPL and look at the 
resulting byte code eg: 

```
object A { val X = "Hello" }
:javap -filter -c A
```

vs

```
object A { final val X = "Hello" }
:javap -filter -c 
```

Also calls to these identifiers (final val) are inlined in the call graph eg:

```
scala> object A { final val X = "Hello" }
scala> object B { final val Z = A.X }
scala> :javap -filter -c B
```

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
dnvriend: I'm for the configuration solution. Please use the strategy used in the akka-persistence-inmemory plugin that uses a configuration called `write-plugin` as implemented in this commit: https://github.com/dnvriend/akka-persistence-inmemory/commit/04884163cdad89860039055766084365b91aa53e#diff-ecbc52c290666a9af5901127454d2bb8R59

src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
aenevala: Cool, that I didn't know. However I'll remove the companion object as I'll switch to use configuration instead. But thanks for the information.

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
aenevala: Ok, sounds good.

aenevala:
I'm done with the changes. Changed events and adapters to look similar as in inmemory-journal tests. Please re-review.

gavares:
@aenevala thanks for doing this. My team is also in need of this change.

dnvriend:
@aenevala thanks, its looking great. will merge it shortly

dnvriend:
@aenevala could you please do the same for the 2.5.0.0 branch. That case the two branches are in sync and have read adapters. Thanks!

PR #104, opened 2017-05-18
Fixed messageQuery method. Filtering messages by deleted flag.

morozov1:
Method _messagesQuery from ReadJournalQueries class did not filter messages by deleted flag.


WellingR:
I checked the akka documentation and from the warning in the message deletion section on http://doc.akka.io/docs/akka/current/scala/persistence.html#message-deletion I conclude that journal plugins may or may not include deleted messages in the queries.

Since the deleted messages are still available, it could make sense to include these deleted messages in the results.

As for this PR, the deleted flag is only checked for the messages query, but not for the events by tag query. I think we should be consistent (either all queries include deleted message, or no queries include deleted messages.

morozov1:
Default implementation of akka persistence with a leveldb plugin does not include deleted events in eventsByPersistenceId query. I think that compatibility is needed.

renatocaval:
I personally think that the idea of deleted messages is flawed since now we have the possibility to query events in situations other of replying events on the PersistentActor.

Why would someone want to delete events? The existing functionality was meant to delete events after taking a snapshot. The goal is to reduce the size of the journal. In which case, only a hard deletion would make sense. 

However, with the introduction of akka-persistence-query, a hard deletion would break it because we may delete events on the command-side, before it gets the chance to be propagated to the query-side. Also, queries added in the future will be totally broken.

The only way of solving that is to have soft-deletion and include the "deleted" events on the queries. But then the question is, why would we want to delete them if we still need them?

That said, I think we should offer the possibility of a hard delete instead, via a configuration flag. 

When using soft deletes, messages are delivered on query side. Kind of useless deletion, but backwards compatible with existing code.

For those really willing to delete events and deal with the impact of this on the query side, we offer a hard delete through some config flag. 

What do you guys think about it?

/cc @morozov1  @WellingR @TimMoore @ignasi35

renatocaval:
@morozov1 I fixed the conflict in this branch and will implement it further as discussed here 
https://gitter.im/akka-persistence-jdbc/contributors?at=5a586129ce68c3bc74aabb85

I will keep the work on this PR so you get the credits as the initiator. If you don't mind. :-)

renatocaval:
I'm closing this on and sending a new PR #162.

This PR was quite old and because of the rebase I did, it became too huge with lots of unrelated commits that make it hard to review.

PR #117, opened 2017-08-24
H2 :: Fix schema name quoting

frederic-gendebien:
I continued trying to work with H2 and I met an error saying that table "snapshot" did not exist. So I investigate and figured out that double quotes were missing into the h2-schema.sql.

I fixed it and now it works fine.


WellingR:
Thanks very much!

PR #118, opened 2017-08-24
upgrades Akka, Slick and Hikari

rcavalcanti:


version.sbt
```diff
@@ -1 +1 @@
-version := "2.5.2.0"
\ No newline at end of file
+version := "2.5.4.0"
```
rcavalcanti: I'm assuming that the convention is that the first three numbers matches the akka version.

WellingR:
Thank you very much for your contribution.

I have just merged my local branch to master. The master branch is now the akka 2.5.x branch and we will maintain a separate akka24 branch.

(side note, the hikari dependency version is declared, but hikaricp is actually a transitive dependency of slick-hikaricp, the dependency version has been removed from the sbt file)

I will close this PR. If you think I missed out anything then please let me know!

rcavalcanti:
Fair enough. Thanks

Issue #116, opened 2017-08-23
Cannot use H2


frederic-gendebien:
GIVEN a project which I just want to use H2 for
WHEN I copy the the content of the configuration you give as an example "h2-application.conf" into my application.conf
THEN I have this issue:
...
Caused by: java.lang.RuntimeException: Failed to load class of driverClassName org.postgresql.Driver
	at com.zaxxer.hikari.HikariConfig.setDriverClassName(HikariConfig.java:323)
	at slick.jdbc.hikaricp.HikariCPJdbcDataSource$.$anonfun$forConfig$3(HikariCPJdbcDataSource.scala:35)
	at slick.jdbc.hikaricp.HikariCPJdbcDataSource$.$anonfun$forConfig$3$adapted(HikariCPJdbcDataSource.scala:35)
	at scala.Option.map(Option.scala:146)
	at slick.jdbc.hikaricp.HikariCPJdbcDataSource$.forConfig(HikariCPJdbcDataSource.scala:35)
	at slick.jdbc.hikaricp.HikariCPJdbcDataSource$.forConfig(HikariCPJdbcDataSource.scala:25)
	at slick.jdbc.JdbcDataSource$.forConfig(JdbcDataSource.scala:48)
	at slick.jdbc.JdbcBackend$DatabaseFactoryDef.forConfig(JdbcBackend.scala:288)
	at slick.jdbc.JdbcBackend$DatabaseFactoryDef.forConfig$(JdbcBackend.scala:285)
	at slick.jdbc.JdbcBackend$$anon$3.forConfig(JdbcBackend.scala:33)
	at akka.persistence.jdbc.util.SlickDatabase$.$anonfun$forConfig$4(SlickDriver.scala:36)
	at scala.Option.getOrElse(Option.scala:121)
	at akka.persistence.jdbc.util.SlickDatabase$.forConfig(SlickDriver.scala:36)
	at akka.persistence.jdbc.journal.JdbcAsyncWriteJournal.<init>(JdbcAsyncWriteJournal.scala:42)
	... 16 more
Caused by: java.lang.ClassNotFoundException: org.postgresql.Driver
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
[...]

PR #120, opened 2017-08-24
Removes Slick profile defautls

rcavalcanti:


README.md
```diff
@@ -42,7 +42,7 @@ Configure `akka-persistence`:
 - instruct akka persistence to use the `jdbc-snapshot-store` plugin,
 
 Configure `slick`:
-- The following slick drivers are supported:
+- The following slick profiles are supported:
```
rcavalcanti: this and subsequent references are point to `profiles` not `drivers`

src/main/resources/reference.conf
```diff
@@ -46,7 +46,8 @@ jdbc-journal {
   parallelism = 8
 
   slick {
-    driver = "slick.jdbc.PostgresProfile$"
+    profile = "slick.jdbc.PostgresProfile$"
+    profile = ${?jdbc-journal.slick.driver}
```
rcavalcanti: This will make sure that an user that have declared the `driver`, but not the `profile` will have his driver config used as a `profile`

src/main/resources/reference.conf
```diff
@@ -103,7 +104,8 @@ jdbc-snapshot-store {
   dao = "akka.persistence.jdbc.snapshot.dao.ByteArraySnapshotDao"
 
   slick {
-    driver = "slick.jdbc.PostgresProfile$"
+    profile = "slick.jdbc.PostgresProfile$"
+    profile = ${?jdbc-snapshot-store.slick.driver}
```
rcavalcanti: same as above

src/main/resources/reference.conf
```diff
@@ -191,7 +193,8 @@ jdbc-read-journal {
   tagSeparator = ","
 
   slick {
-    driver = "slick.jdbc.PostgresProfile$"
+    profile = "slick.jdbc.PostgresProfile$"
+    profile = ${?jdbc-read-journal.slick.driver}
```
rcavalcanti: same as above

WellingR:
Hey @rcavalcanti thanks for your contribution. I have actually already included a part of these changes already. However your addition of `profile = ${?jdbc-journal.slick.driver}` is a good one.

Could you rebase your branch against the latest master and resolve the conflict? Then I will merge this PR

rcavalcanti:
@WellingR rebased and conflict resolved

src/main/resources/reference.conf
```diff
@@ -47,6 +47,7 @@ jdbc-journal {
 
   slick {
     profile = "slick.jdbc.PostgresProfile$"
+    profile = ${?jdbc-journal.slick.driver}
```
rcavalcanti: I did some local test to simulate the case in which an user have configured `driver` and we kind of move it to `profile`. 

That will only work if `driver = slick.jdbc.PostgresProfile$`.

If the user has it as `driver = slick.driver.PostgresDriver$`, we are in trouble. Because Slick will see it as the value of `profile` property and try to load `slick.driver.PostgresDriver`. And that will fail because there is not such a class anymore, only a type alias.

WellingR: I see your point. In that case it would actually be better to not include this setting.

In the end it might actually make sense to remove all references to postgress at all. In this case a misconfigured database configuration would fail to load (hopefully with a useful error message). Removing this default would actually prevent issues such as #116

rcavalcanti: It's true that the `reference.conf` should not contain any settings that are specific to environment. I think that it's often used as a reference for documentation, but that's not the best thing to do because we end up setting default values that can be very confusing.

rcavalcanti:
@WellingR I pushed a workaround for it. A ugly one. 
Not yet sure we must take that path, but let's discuss it next week. I need a fresh mind on Monday to evaluate it. 

Eventually, this is something we should try to solve in Slick itself.

We are facing this same issue in Lagom. The reason for that is because here and there (Lagom) we are providing default drivers in a `reference.conf` file. That's great for quick bootstrap, but now that we want / need to promote `slick.profile` over `slick.driver` we run the risk of overriding properties in production system.

WellingR:
A potential solution would be to comment the lines specifying the profile, driver (and perhaps also the other environment related settings such as url, user, password etc).

These commented lines serve as documentation in the reference conf file. But since we cannot provide sensible defaults, it it actually better to not provide these defaults at all.

I ran some of the tests where I removed the profile, which gives a good error message. The same holds if the url is not specified. Somehow if the jdbc driver is not specified the tests seem to pass, maybe slick determines this based on the url.

@rcavalcanti what do you think about this?

rcavalcanti:
@WellingR I also think this is the best to do. I will adapt this PR accordingly and remove the workaround I added. 

For the test, it may be that the are overriding those values which would explain why it still works. In any case, the test should define their own values and production code must not define defaults.

rcavalcanti:
@WellingR I updated this PR. 

- removed workaround added ealier
- commented out the database specific settings in `reference.conf`
- removed references to boot2docker and docker (doesn't make sense to have it there)
- minor tweaks in comments in `reference.conf`

rcavalcanti:
BTW, this PR can be easily cherrypicked into akka-2.4 branch

PR #127, opened 2017-09-20
fixes for max ordering jump

### LINKED ISSUE #126: Max ordering can jump back to a previous value
In `JournalSequenceActor` we have two competing queries to determine the max ordering. 

When `QueryOrderingIds` takes time to stabilise, the value retrieved by `ScheduleAssumeMaxOrderingId` will become the new `currentMaxOrdering`.

The following scenario may occur:
1) `currentMaxOrdering = 50`, `QueryOrderingIds` fetches orderings starting from 50
2) `ScheduleAssumeMaxOrderingId` updates `currentMaxOrdering` to 300
3) `NewOrderingIds` arrive with a seq of orderings from 50 to `batchSize`.

On [`JournalSequenceActor.scala:84`](https://github.com/dnvriend/akka-persistence-jdbc/blob/master/src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala#L84) we decide what will be the `newMax`:
```
val newMax =
  if (currentMax + 1 == elem) elem
  else if ((currentMax + 1).until(elem).forall(givenUp.contains)) elem
  else currentMax
```

The `else if` branch will be executed an the `newMax` will be equal to `elem`.
* `currentMax = 300`
* `(currentMax + 1).until(elem)` == `301.until(50)` == empty range
* range.forall(givenUp.contains)) always returns true on an empty range. 
* as result, the `nextMax` will become 50. So, we spring back to 50 while we were already on 300.

rcavalcanti:
This PR solves the case where the result of `ScheduleAssumeMaxOrderingId` could be ignored when the result from `QueryOrderingIds` is processed

(fixes #126)

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
+++ b/src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
@@ -117,7 +117,7 @@ class JournalSequenceActor(readJournalDao: ReadJournalDao, config: JournalSequen
 
             // if it's a gap and has been detected before on a previous iteration we give up
             // that means that we consider it a genuine gap that will never be filled
-            case e if givenUp(e) => missing
+            case e if givenUp(e)               => missing
 
             // any other case is a gap that we expect to be filled soon
             case _ =>
@@ -135,7 +135,12 @@ class JournalSequenceActor(readJournalDao: ReadJournalDao, config: JournalSequen
           (newMax, currentElement, newMissing)
       }
 
-    val newMissingByCounter = missingByCounter + (moduloCounter -> missingElems)
+    val newMissingByCounter =
+      (missingByCounter + (moduloCounter -> missingElems))
+        .map {
+          case (key, value) =>
+            key -> value.filter(missingId => missingId > nextMax)
+        }
 
     // did we detect gaps in the current batch?
     val noGapsFound = newMissingByCounter.values.forall(_.isEmpty)
--- a/src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
+++ b/src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
@@ -13,7 +13,7 @@ object JournalSequenceActor {
   def props(readJournalDao: ReadJournalDao, config: JournalSequenceRetrievalConfig)(implicit materializer: Materializer): Props = Props(new JournalSequenceActor(readJournalDao, config))
 
   private case object QueryOrderingIds
-  private case class NewOrderingIds(elements: Seq[OrderingId])
+  private case class NewOrderingIds(originalOffset: Long, elements: Seq[OrderingId])
 
   private case class ScheduleAssumeMaxOrderingId(max: OrderingId)
   private case class AssumeMaxOrderingId(max: OrderingId)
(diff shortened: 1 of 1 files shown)


rcavalcanti:
@WellingR In this PR I also added documentation to try to explain what each piece is doing. 

The code is not trivial and I found it quite challenging to following. So I added some explanations. There are some parts that become obsolete. I also tried to remove double negations. See my comments in the PR.

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
```diff
@@ -57,47 +57,34 @@ class JournalSequenceActor(readJournalDao: ReadJournalDao, config: JournalSequen
...
-      }
+      readJournalDao
+        .journalSequence(currentMaxOrdering, batchSize)
+        .runWith(Sink.seq)
+        .map(result => NewOrderingIds(currentMaxOrdering, result)) pipeTo self
+
+    case NewOrderingIds(originalOffset, _) if originalOffset < currentMaxOrdering =>
+      // search was done using an offset that became obsolete in the meantime
+      // therefore we start a new query
+      self ! QueryOrderingIds
```
rcavalcanti: This cover the case when the `currentMaxOrdering` has been updated by `ScheduleAssumeMaxOrderingId`

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
```diff
@@ -57,47 +57,34 @@ class JournalSequenceActor(readJournalDao: ReadJournalDao, config: JournalSequen
...
+
+    case NewOrderingIds(_, elements) =>
+      findGaps(elements, currentMaxOrdering, missingByCounter, moduloCounter)
```
rcavalcanti: I found better to move the main algorithm code to it's one method. That gives better visibility when we want to have a glimpse on all the messages that the actor can handle.

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
```diff
@@ -107,6 +94,69 @@ class JournalSequenceActor(readJournalDao: ReadJournalDao, config: JournalSequen
...
+            // that means that we consider it a genuine gap that will never be filled
+            case e if givenUp(e) => missing
+
+            // any other case is a gap that we expect to be filled soon
+            case _ =>
+              val currentlyMissing = previousElement + 1 until currentElement
+              // we don't want to declare it as missing if it has been already declared on a previous iterations
+              def alreadyMissing(e: Long) = missingByCounter.values.exists(_.contains(e))
+              missing ++ currentlyMissing.filterNot(alreadyMissing)
+          }
```
rcavalcanti: Choose to use pattern matching because is helps writing explanation for each case.
(also removed double negations)

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
```diff
@@ -107,6 +94,69 @@ class JournalSequenceActor(readJournalDao: ReadJournalDao, config: JournalSequen
...
+
+          // we must decide if the if we move the cursor forward
+          val newMax =
+            if ((currentMax + 1).until(currentElement).forall(givenUp.contains)) currentElement
+            else currentMax
```
rcavalcanti: Previous check was:
```scala
if (currentMax + 1 == elem) elem
else if ((currentMax + 1).until(elem).forall(givenUp.contains)) elem		
else currentMax
```

The first branch is unnecessary because when `currentMax + 1 == elem`, the range will be empty and the second branch will always return true.

```scala
val a = 1
val b = 2
(a +1).until(b).forall(_ => false)
res2: Boolean = true
```
`forall` on an empty range always return true.

The only drawback of removing the first if is that it becomes less clear. Maybe we should put it back for the sake of clarity. Any thoughts?

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
```diff
@@ -107,6 +94,69 @@ class JournalSequenceActor(readJournalDao: ReadJournalDao, config: JournalSequen
...
+      }
+
+    val newMissingByCounter = missingByCounter + (moduloCounter -> missingElems)
+
+    // did we detect gaps in the current batch?
+    val noGapsFound = newMissingByCounter.values.forall(_.isEmpty)
+
+    // full batch means that we retrieved as much elements as the batchSize
+    // that happens when we are not yet at the end of the stream
+    val isFullBatch = elements.size >= batchSize
```
rcavalcanti: This was previously `nextmax - currentMaxOrdering >= batchSize`. 
It took me some time to understand the motivation. 

Am I correct in interpreting it that way?

rcavalcanti:
btw, I didn't add any extra test for this. The reason is that what it's fixing is invisible. When the 'jump' occurs, nothing bad happens. It only neutralises the work done by `ScheduleAssumeMaxOrderingId`

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
```diff
@@ -107,6 +94,69 @@ class JournalSequenceActor(readJournalDao: ReadJournalDao, config: JournalSequen
       context.become(receive(currentMaxOrdering, missingByCounter, moduloCounter, newDelay))
   }
 
+  /**
+   * This method that implements the "find gaps" algo. It's the meat and main purpose of this actor.
+   */
+  def findGaps(elements: Seq[OrderingId], currentMaxOrdering: OrderingId, missingByCounter: Map[Int, Set[OrderingId]], moduloCounter: Int) = {
+
+    // list of elements that will be considered as genuine gaps.
+    // `givenUp` is whether empty or is was filled on a previous iteration
```
WellingR: typo: whether -> either

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
```diff
@@ -107,6 +94,69 @@ class JournalSequenceActor(readJournalDao: ReadJournalDao, config: JournalSequen
...
+
+            // any other case is a gap that we expect to be filled soon
+            case _ =>
+              val currentlyMissing = previousElement + 1 until currentElement
+              // we don't want to declare it as missing if it has been already declared on a previous iterations
+              def alreadyMissing(e: Long) = missingByCounter.values.exists(_.contains(e))
+              missing ++ currentlyMissing.filterNot(alreadyMissing)
+          }
+
+          // we must decide if the if we move the cursor forward
```
WellingR: typo: `if the if`

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
WellingR: Looks good to me, you can add a comment for clarity if you want

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
WellingR: Yes you are correct, using elements.size is indeed easier to understand

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
```diff
@@ -107,6 +94,69 @@ class JournalSequenceActor(readJournalDao: ReadJournalDao, config: JournalSequen
...
+    // that happens when we are not yet at the end of the stream
+    val isFullBatch = elements.size >= batchSize
+
+    if (noGapsFound && isFullBatch) {
+      // Many elements have been retrieved but none are missing
+      // We can query again immediately, as this allows the actor to rapidly retrieve the real max ordering
+      self ! QueryOrderingIds
+      context.become(receive(nextMax, newMissingByCounter, moduloCounter))
+    } else {
+      // whether we detected gaps or we reached the end of stream (batch not full)
```
WellingR: wheter -> either

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
WellingR: Looks good.

Perhaps it would be a good idea to move the `context.become(receive(...))` out of the findGaps method and into this receive function again. That makes is more clear that we are changing the state of the actor.

rcavalcanti: I agree, that would be much clear if all `context.become` appear on the same level.

rcavalcanti: I just tried and unfortunately it depends on 3 values that are only in the scope of `findGaps`.
* `noGapsFound && isFullBatch` - Boolean (two values but reducible to 1)
* `nextMax` - Long
* `newMissingByCounter` - Map[Int, Set[JournalSequenceActor.OrderingId]]

Not sure if returning a case class or a tuple3 (Boolean, Long, Map[Int, Set[JournalSequenceActor.OrderingId]]) will add clarity to it.

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
rcavalcanti: silly typo, thanks

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
rcavalcanti: if it happens twice is not silly anymore, it's stupid. :-(
Thanks again

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
rcavalcanti: BTW, just realised that it should be `elements.size == batchSize`, not `>=`

PR #129, opened 2017-09-30
currentEventsByTag fix ( #124 )

### LINKED ISSUE #124: CurrentEventsByTag potentially broken
`currentEventsByTag` is supposed to return a finite `Source` that starts from passed `Offset` up to last known event. When it finishes, it finishes and new events are not delivered. It's possible that such a query returns an empty `Source`. 

It relies on the `JournalSequenceActor` to determine what is the max known ordering, but just after starting, that actor considers that the max ordering is 0. It takes a while before it "learns" what is the real max ordering. During that period, `currentEventsByTag` will return an empty `Source` even if there are events starting from the passed offset.

This won't happen only at start up time. A system with a journal with tens of thousands events will also suffer from it. 

For instance, the default buffer size is 10000. For a journal with 100k events, a call to `currentEventsByTag` starting from offset 30000, will return an empty `Source` in the beginning. After some time, when the max ordering becomes, let's say 50000, it will return a `Source` from 30001 to 50000, while the correct one would have been 30001 to 100000.

WellingR:
This is should fix #124.

This PR is ready be merged, however in the future we might want to refactor some of the code in the JdbcReadJournal to make it easier to test.

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
+++ b/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
@@ -45,16 +45,16 @@ object JdbcReadJournal {
   private sealed trait FlowControl
 
   /** Keep querying - used when we are sure that there is more events to fetch */
-  private object Continue extends FlowControl
+  private case object Continue extends FlowControl
 
   /**
    * Keep querying with delay - used when we have consumed all events,
    * but want to poll for future events
    */
-  private object ContinueDelayed extends FlowControl
+  private case object ContinueDelayed extends FlowControl
 
   /** Stop querying - used when we reach the desired offset  */
-  private object Stop extends FlowControl
+  private case object Stop extends FlowControl
 }
 
 class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem) extends ReadJournal
@@ -188,6 +188,8 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem) 
               terminateAfterOffset match {
                 // we may stop if target is behind queryUntil and we don't have more events to fetch
                 case Some(target) if !hasMoreEvents && target <= queryUntil.maxOrdering => Stop
+                // We may also stop if we have found an event with an offset >= target
+                case Some(target) if xs.exists(_.offset.value >= target) => Stop
 
                 // otherwise, disregarding if Some or None, we must decide how to continue
                 case _ =>
--- a/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
+++ b/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
@@ -19,7 +19,6 @@ package scaladsl
 
 import akka.NotUsed
 import akka.actor.ExtendedActorSystem
-import akka.persistence.jdbc.JournalRow
 import akka.persistence.jdbc.config.ReadJournalConfig
 import akka.persistence.jdbc.query.JournalSequenceActor.{GetMaxOrderingId, MaxOrderingId}
 import akka.persistence.jdbc.query.dao.ReadJournalDao
@@ -79,7 +78,7 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem) 
   }
 
   // Started lazily to prevent the actor for querying the db if no eventsByTag queries are used
-  private[query] lazy val orderingActor = system.actorOf(JournalSequenceActor.props(readJournalDao, readJournalConfig.journalSequenceRetrievalConfiguration))
+  private[query] lazy val journalSequenceActor = system.actorOf(JournalSequenceActor.props(readJournalDao, readJournalConfig.journalSequenceRetrievalConfiguration))
   private val delaySource =
     Source.tick(readJournalConfig.refreshInterval, 0.seconds, 0).take(1)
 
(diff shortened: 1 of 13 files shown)


src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -135,39 +134,63 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem)
...
+            val nextStartingOffset = if (xs.isEmpty) from else xs.map(_.offset.value).max
+            Some((nextStartingOffset, completeStreamAfterThisBatch, isFullBatch), xs)
+          }
+        }
+        if (completeImmediately) Future.successful(None)
+        else {
+          if (retrieveImmediately) retrieveNextBatch()
+          else akka.pattern.after(readJournalConfig.refreshInterval, system.scheduler)(retrieveNextBatch())
+        }
+    }.mapConcat(identity)
```
rcavalcanti: That's a more general comment. 

I find it hard to reason when using two booleans to express three states. IIUC, we need here a flow control with three states: `Continue`, `ContinueDelayed` and `Stop`. 
 
I suggest the following:
```scala

sealed trait FlowControl
    
/** Keep querying - used when we are sure that there is more events to fetch */
object Continue extends FlowControl

/** Keep querying with delay - used when we have consumed all events,
 * but want to poll for future events */
object ContinueDelayed extends FlowControl

/** Stop querying - used when we reach the desired offset  */
object Stop extends FlowControl


Source.unfoldAsync[(Long, FlowControl), Seq[EventEnvelope]]((offset, Continue)) {

  case (from, control) =>

    val batchSize = readJournalConfig.maxBufferSize

    def retrieveNextBatch() = {

      for {
        xs <- currentJournalEventsByTag(tag, from, batchSize, queryUntil).runWith(Sink.seq)
      } yield {

        def decideHowToContinue = {
          val rangeSize = queryUntil - from
          val expectedBatchSize = if (rangeSize < batchSize) rangeSize else batchSize

          // if we fetch less than expected it means we reached the end of the stream
          // we want to keep querying, but adding some delay
          if (xs.size < expectedBatchSize) ContinueDelayed
          else Continue
        }

        val control =
          terminateAfterOffset
            // keep it as Some if target is reached
            .filterNot(_ > queryUntil.maxOrdering)
            // we can stop because we have reached the target offset
            .map(_ => Stop)
            // othewise decide how to continue
            .getOrElse(decideHowToContinue)

        val nextStartingOffset = if (xs.isEmpty) from else xs.map(_.offset.value).max

        Some((nextStartingOffset, control), xs)
      }
    }

    control match {
      case Stop => Future.successful(None)
      case Continue => retrieveNextBatch()
      case ContinueDelayed =>
        akka.pattern.after(readJournalConfig.refreshInterval, system.scheduler)(retrieveNextBatch())
    }

}.mapConcat(identity)
```

NOTE: this implementation resolves the points I mentioned before:
 * start without delay
 * isFullBatch fix
 * stop as soon as we detected we reach the target

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -135,39 +134,63 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem)
...
+     */
+    Source.unfoldAsync[(Long, Boolean, Boolean), Seq[EventEnvelope]]((offset, false, false)) {
+      case (from: Long, completeImmediately: Boolean, retrieveImmediately: Boolean) =>
+        def retrieveNextBatch(): Future[Some[((Long, Boolean, Boolean), Seq[EventEnvelope])]] = {
+          for {
+            queryUntil <- journalSequenceActor.ask(GetMaxOrderingId).mapTo[MaxOrderingId]
+            xs <- currentJournalEventsByTag(tag, from, readJournalConfig.maxBufferSize, queryUntil).runWith(Sink.seq)
+          } yield {
+            val isFullBatch = xs.size == readJournalConfig.maxBufferSize
+            val completeStreamAfterThisBatch = !isFullBatch && terminateAfterOffset.exists(_ <= queryUntil.maxOrdering)
```
rcavalcanti: I don't think it's need to check if batch is full. If the target offset is already reached we are safe to say that we don't need another iteration.

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -135,39 +134,63 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem)
...
+     *   executed immediately, or after a delay.
+     */
+    Source.unfoldAsync[(Long, Boolean, Boolean), Seq[EventEnvelope]]((offset, false, false)) {
+      case (from: Long, completeImmediately: Boolean, retrieveImmediately: Boolean) =>
+        def retrieveNextBatch(): Future[Some[((Long, Boolean, Boolean), Seq[EventEnvelope])]] = {
+          for {
+            queryUntil <- journalSequenceActor.ask(GetMaxOrderingId).mapTo[MaxOrderingId]
+            xs <- currentJournalEventsByTag(tag, from, readJournalConfig.maxBufferSize, queryUntil).runWith(Sink.seq)
+          } yield {
+            val isFullBatch = xs.size == readJournalConfig.maxBufferSize
```
rcavalcanti: @WellingR, I think I have spotted a small bug here. 

The `isFullBatch` is flawed. The following scenario may happen:
```
journalSize = 1000
bufferSize = 100
queryFrom = 50 // user provided offset
queryUntil = 100 // max offset as retrieved from journalSequenceActor
```
The size of `xs` will be limited by the `queryFrom` offset. In this example, it will be 50 and it will introduce a unnecessary delay on line 180. 

I will review it in the code I sent on my last PR #127. That's probably the case there as well.

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -135,39 +134,63 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem)
...
-      .mapAsync(1)(Future.fromTry)
+    implicit val askTimeout: Timeout = Timeout(readJournalConfig.journalSequenceRetrievalConfiguration.askTimeout)
+    /* We unfold with 3 parameters:
+     * - The first parameter (from) determines the starting offset (i.e. the minimum value) to query from.
+     * - The second parameter (completeImmediately) signifies that we have queried events until at least
+     *   `terminateAfterOffset`, therefore the stream no longer needs to continue.
+     * - The third parameter (retrieveImmediately) is a boolean which determines whether the database query should be
+     *   executed immediately, or after a delay.
+     */
+    Source.unfoldAsync[(Long, Boolean, Boolean), Seq[EventEnvelope]]((offset, false, false)) {
```
rcavalcanti: We may want to start without introducing a delay, (offset, false, true).

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
WellingR: Actually, maybe just the name if flawed, if the batchSize is equal to readJournalConfig.maxBufferSize then we can query again immediately.

if the batch size is smaller because `queryUntil` was somewhat small, then we also need to wait because the journal Sequence actor needs some time to increment the max ordering id

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
WellingR: Looks good. However I think we should introduce a in case the size of the batch is range size (as I mentioned in my other comment)

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
WellingR: We do need this check.

Say we query the journal sequence actor, it states that the maxOrdering id is 1000. We then execute with the following parameters:
```
bufferSize = 100
queryFrom = 50 // user provided offset
queryUntil = 1000 // max offset as retrieved from journalSequenceActor
```
We receive events 51 to 151. The pitfall here is that `queryUntil` is a max used in the query, but that does not mean that we actually receive all events until queryUntil

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
rcavalcanti: > Actually, maybe just the name if flawed, if the batchSize is equal to readJournalConfig.maxBufferSize then we can query again immediately.

Maybe not, if the `queryUntil` is the last item in the batch, we may have reached the end of it, for example:
```
bufferSize = 100
queryFrom = 50 // user provided offset
queryUntil = 151 
fetchde = 51 to 151
```
But I guess this is a minor issue. Will be sorted out on next iteration.

What about that:

```scala
def decideHowToContinue = {
  val possibleRangeSize = queryUntil.maxOrdering - from
  if (possibleRangeSize > xs.size || xs.size == batchSize) {
    // there are certainly more to fetch
    Continue
  } else {
    // if query is not being limited by batch size and we fetched all possible items,
    // we can consider that there is nothing else to fetch for the moment
    ContinueDelayed
  }
}
```

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
rcavalcanti: Yes, you are right. I didn't consider that case.

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
rcavalcanti: Maybe to make clear that we have indeed fetched the targeted offset...
```scala

val offsetsOnly = xs.map(_.offset.value)
val control =
  terminateAfterOffset
    // keep it as Some if target is reached
    .filter(target => offsetsOnly.contains(target))
    // we can stop because we have reached the target offset
    .map(_ => Stop)
    // otherwise we must decide how to continue
    .getOrElse(decideHowToContinue)
```

PR #131, opened 2017-10-06
Current events by tag fix wtih flow control

rcavalcanti:
This PR is based on previous PR from @WellingR (PR #129). 

It introduces a `FlowControl` model that replaces the double boolean logic by a three state ADT.

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
+++ b/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
@@ -187,9 +187,8 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem) 
 
             // calculates if we have more events to fetch
             val hasMoreEvents = {
-
               val possibleRangeSize = queryUntil.maxOrdering - from
-
+              
               // when batch is being limited by queryUntil
               // we can consider that there is nothing else to fetch
               if (possibleRangeSize <= batchSize) false
@@ -205,7 +204,8 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem) 
                 // otherwise, disregarding if Some or None,
                 // we must decide how to continue
                 case _ =>
-                  // continue immediately when retrieving full batches
+                  // continue immediately when are sure that there are more events to be fetched
+                  // otherwise let it happen within configured delay
                   if (hasMoreEvents) Continue else ContinueDelayed
               }
 
--- a/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
+++ b/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
@@ -173,45 +173,28 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem) 
     import akka.pattern.ask
     import JdbcReadJournal._
     implicit val askTimeout: Timeout = Timeout(readJournalConfig.journalSequenceRetrievalConfiguration.askTimeout)
+    val batchSize = readJournalConfig.maxBufferSize
 
     Source.unfoldAsync[(Long, FlowControl), Seq[EventEnvelope]]((offset, Continue)) {
-
       case (from, control) =>
-
-        val batchSize = readJournalConfig.maxBufferSize
-
         def retrieveNextBatch() = {
           for {
             queryUntil <- journalSequenceActor.ask(GetMaxOrderingId).mapTo[MaxOrderingId]
             xs <- currentJournalEventsByTag(tag, from, batchSize, queryUntil).runWith(Sink.seq)
           } yield {
 
-            // calculates if we have more events to fetch
-            val hasMoreEvents = {
-              val possibleRangeSize = queryUntil.maxOrdering - from
-              
-              // when batch is being limited by queryUntil
-              // we can consider that there is nothing else to fetch
-              if (possibleRangeSize <= batchSize) false
-              // otherwise, base decision solely on batch size
-              else xs.size == batchSize
-            }
-
+            val hasMoreEvents = xs.size == batchSize
             val control =
               terminateAfterOffset match {
                 // we may stop if target is behind queryUntil and we don't have more events to fetch
                 case Some(target) if !hasMoreEvents && target <= queryUntil.maxOrdering => Stop
 
-                // otherwise, disregarding if Some or None,
-                // we must decide how to continue
+                // otherwise, disregarding if Some or None, we must decide how to continue
                 case _ =>
-                  // continue immediately when are sure that there are more events to be fetched
-                  // otherwise let it happen within configured delay
                   if (hasMoreEvents) Continue else ContinueDelayed
               }
 
             val nextStartingOffset = if (xs.isEmpty) from else xs.map(_.offset.value).max
-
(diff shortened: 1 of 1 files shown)


src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -135,39 +136,96 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem)
...
+
+          for {
+            queryUntil <- journalSequenceActor.ask(GetMaxOrderingId).mapTo[MaxOrderingId]
+            xs <- currentJournalEventsByTag(tag, from, batchSize, queryUntil).runWith(Sink.seq)
+          } yield {
+
+            val offsetsOnly = xs.map(_.offset.value)
+
+            val control =
+              terminateAfterOffset
```
WellingR: A simple pattern match on `terminateAfterOffset` might make this easier to read

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -135,39 +136,96 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem)
...
+                // keep it as Some if target is reached
+                .filter(target => offsetsOnly.contains(target))
```
WellingR: It could be that the `target` does not have the requested tag at all. In this case we will never terminate.
If first hand you might think that the largest offset (or the value `nextStartingOffset` as defined below) could be used for comparison. But it cannot.

Consider the following case: (actually we should add a test for this).
A currentEventsByTag query for a tag which does not exist in the journal should complete (without returning any events).

I believe that with the current code, the query will not return any events, but the stream will also never complete.

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
rcavalcanti: True. I had another version using `collect`, but I wasn't convinced of it. 

I will try a simple pattern match as suggested

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -135,39 +136,104 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem)
...
+            val hasMoreEvents = {
+
+              val possibleRangeSize = queryUntil.maxOrdering - from
+
+              // when batch is being limited by queryUntil and we did fetch everything we could,
+              // we can consider that there is nothing else to fetch
+              if (possibleRangeSize <= batchSize && xs.size == possibleRangeSize) false
+              // otherwise, base decision solely on batch size
+              else xs.size == batchSize
+            }
```
rcavalcanti: The `hasMoreEvents` tried to identify we there are more events to fetch. 

It covers the following cases:

### Scenario 1
Batch was limited by `queryUntil` making it smaller than configured `batchSize`. 
That means that fetched everything we could and we can consider

### Scenario 2
Batch was limited by the batch size and we fetched as much as batch size. 
In that case we can't know if there are more events or not, so we assume there are.

### Scenario 3
Batch was limited by the batch size and we fetched less items.
In that case can consider that we reached the end of the stream.

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -139,46 +141,87 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem)
...
           for {
             queryUntil <- journalSequenceActor.ask(GetMaxOrderingId).mapTo[MaxOrderingId]
-            xs <- currentJournalEventsByTag(tag, from, readJournalConfig.maxBufferSize, queryUntil).runWith(Sink.seq)
+            xs <- currentJournalEventsByTag(tag, from, batchSize, queryUntil).runWith(Sink.seq)
           } yield {
-            val isFullBatch = xs.size == readJournalConfig.maxBufferSize
-            val completeStreamAfterThisBatch = !isFullBatch && terminateAfterOffset.exists(_ <= queryUntil.maxOrdering)
+
+            // calculates if we have more events to fetch
+            val hasMoreEvents = {
```
WellingR: We can simplify this to `val hasMoreEvents = xs.size == batchSize`

The only slight difference is the case where the possible size range is equal to batchSize. However I personally favour simple and understandable code above complex codes that handles every edge case with the best efficiency.

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -139,46 +141,87 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem)
...
     import akka.pattern.ask
     implicit val askTimeout: Timeout = Timeout(readJournalConfig.journalSequenceRetrievalConfiguration.askTimeout)
-    /* We unfold with 3 parameters:
-     * - The first parameter (from) determines the starting offset (i.e. the minimum value) to query from.
-     * - The second parameter (completeImmediately) signifies that we have queried events until at least
-     *   `terminateAfterOffset`, therefore the stream no longer needs to continue.
-     * - The third parameter (retrieveImmediately) is a boolean which determines whether the database query should be
-     *   executed immediately, or after a delay.
+
+    sealed trait FlowControl
```
WellingR: To make this method a bit shorter, we could probably move this sealed trait to the compation object of this JdbcReadJournal (then we should probably also make it private since it it not meant to be used outside)

WellingR:
General comments:

- Try to keep whitespace changes to a minimum, which makes reviewing easier
- Codacity is complaining about method length. I already made the suggestion to move the flowcontrol trait outside of the method. For the rest, the method contains lots of white lines, and lots of comments. Some of these comments may not be needed: e.g. `hasMoreEvents` already has a descriptive name, so no explanation is needed for that :)

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
rcavalcanti: I will move back to simply `xs.size == bacthSize`. 

My point was not to cover all edge cases, but to make it understandable. For instance, `val hasMoreEvents = xs.size == batchSize` doesn't convey the mental work we did to come to the conclusion that there is more events, or not. 

But I agree that my trial is not doing it better neither.

rcavalcanti:
I will move the traits to companion object. That's a good hint. 

About formatting and comments, I will try to reduce it, but I want to make two arguments in favour of both:

#### About comments
This kind of code is very dense. There is some, sometimes non-trivial, mental work in there and without enough comments we will make it hard to maintain it. 

#### About break-lines
Empty lines increases readability and reduces the cognitive load when parsing code with a human eye. For the same reason we do use punctuation and paragraphs when writing in natural languages. I think that we should not compromise on that because Codacy is saying so.

WellingR:
I agree with your statements about the line breaks and comments. With the improvement you made it looks much and much better.

This can be merged with my PR if you agree

rcavalcanti:
Thanks @WellingR. I'm merging it.

PR #133, opened 2017-10-08
Changed oracle test-docker image

WellingR:
The travis build was broken. Because the `wnameless/oracle-xe-11g` image no longer exists on dockerhub. I chose the most-starred oracle-xe-11g instead, which should resolve our problems.


WellingR:
Note: An issue exists in the git repo of the image we used before: https://github.com/wnameless/docker-oracle-xe-11g/issues/70

PR #136, opened 2017-10-26
currentEventsByTag fix ( #124 ) + exclusive offsets queries

renatocaval:
This PR bring the fix for `currentEventsByTag` and the exclusive offsets queries that was never merged in akka24.


WellingR:
Looks good 👍

PR #145, opened 2017-11-16
Set a name for the journal sequence actor

WellingR:
This ensures better logging in case of issues (e.g. dead letter warnings)


renatocaval:
Backported on akka24 branch (20a8c955a5ad7e4012ae74c1afb9126f8ebd36bc)

Issue #93, opened 2017-03-16
ReadJournalDao directly wired to JournalRow


svenwiegand:
At first: Thanks for that great persistence solution.

I like the possibility to provide custom DAO implementations. But there is one weak point in the `ReadJournalDao`: The `eventsByTag` is expected to return a `JournalRow`:

```scala
def eventsByTag(tag: String, offset: Long, max: Long): Source[Try[(PersistentRepr, Set[String], JournalRow)], NotUsed]
```

This makes it _nearly_ impossible to provide an own `ReadJournalDao` implementation based on something else but the `JournalRow`.

WellingR:
Closing this now that #146 has been merged. The changes will be included in the next release.

PR #146, opened 2017-11-16
Dependency on JournalRow removed from readjournal

### LINKED ISSUE #76: ReadJournalDao is tied to ByteArrayDao
Hi,

I have some trouble to implement my own ReadJournal Dao. Indeed, this latter defines the method eventsByTag with a return type that references akka.persistence.jdbc.dao.bytea.journal.JournalTables.JournalRow....which is the byte-array-based default DAO implementation. 
Hence, it is not possible to implement my own version based on different row type.

From what I understood by searching through the code, this JournalRow value is used only to retrieve the "ordering" value of the journal row.

As a very ugly workaround, I can construct a fake bytea.JournalRow.
I was also thinking about forking JdbcReadJournal

WellingR:
Fixes #76 and #93 

Note this is a breaking change in the api of JdbcReadJournalDao, it should only affect those who actually implement their own read journal.

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalSerializer.scala
+++ b/src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalSerializer.scala
@@ -38,8 +38,8 @@ class ByteArrayJournalSerializer(serialization: Serialization, separator: String
       ))
   }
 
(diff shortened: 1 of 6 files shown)


WellingR:
Yes we can backport this, however this is a breaking change on the DAO api.

PR #150, opened 2017-11-27
[WIP] SlickDriver configures an AsyncExecutor for JNDI resources

### LINKED ISSUE #142: Provide a way to configure maxConnections when using a JNDI name
Slick 3.2 requires knowledge about the maximum number of connections available from a data source. When using an external data source, this value needs to be passed in explicitly (see https://github.com/slick/slick/pull/1685).

Currently, Akka Persistence JDBC [passes `None`](https://github.com/dnvriend/akka-persistence-jdbc/blob/master/src/main/scala/akka/persistence/jdbc/util/SlickDriver.scala#L34) for this value, which is only meant to be use for connection pools with no limit.

renatocaval:
Fixes #142

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
+++ b/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
@@ -93,7 +93,10 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem) 
 
   // Started lazily to prevent the actor for querying the db if no eventsByTag queries are used
   private[query] lazy val journalSequenceActor = system.actorOf(
-    JournalSequenceActor.props(readJournalDao, readJournalConfig.journalSequenceRetrievalConfiguration), "akka-persistence-jdbc-journal-sequence-actor")
+    JournalSequenceActor.props(readJournalDao, readJournalConfig.journalSequenceRetrievalConfiguration),
+    "akka-persistence-jdbc-journal-sequence-actor"
+  )
+
   private val delaySource =
     Source.tick(readJournalConfig.refreshInterval, 0.seconds, 0).take(1)
 
@@ -190,7 +193,7 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem) 
                 // we may stop if target is behind queryUntil and we don't have more events to fetch
                 case Some(target) if !hasMoreEvents && target <= queryUntil.maxOrdering => Stop
                 // We may also stop if we have found an event with an offset >= target
-                case Some(target) if xs.exists(_.offset.value >= target) => Stop
+                case Some(target) if xs.exists(_.offset.value >= target)                => Stop
 
                 // otherwise, disregarding if Some or None, we must decide how to continue
                 case _ =>
--- a/src/main/scala/akka/persistence/jdbc/util/SlickDriver.scala
+++ b/src/main/scala/akka/persistence/jdbc/util/SlickDriver.scala
@@ -17,12 +17,15 @@
 package akka.persistence.jdbc.util
 
 import javax.naming.InitialContext
+import javax.sql.DataSource
 
-import akka.persistence.jdbc.config.SlickConfiguration
+import akka.persistence.jdbc.config.{AsyncExecutorConfig, JournalConfig, SlickConfiguration}
 import com.typesafe.config.Config
+import slick.SlickException
 import slick.basic.DatabaseConfig
 import slick.jdbc.JdbcProfile
 import slick.jdbc.JdbcBackend._
+import slick.util.AsyncExecutor
 
 object SlickDriver {
   def forDriverName(config: Config): JdbcProfile =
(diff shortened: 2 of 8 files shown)


src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -93,7 +93,10 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem)
 
   // Started lazily to prevent the actor for querying the db if no eventsByTag queries are used
   private[query] lazy val journalSequenceActor = system.actorOf(
-    JournalSequenceActor.props(readJournalDao, readJournalConfig.journalSequenceRetrievalConfiguration), "akka-persistence-jdbc-journal-sequence-actor")
+    JournalSequenceActor.props(readJournalDao, readJournalConfig.journalSequenceRetrievalConfiguration),
+    "akka-persistence-jdbc-journal-sequence-actor"
+  )
```
renatocaval: Scalariform was changing this line and doing something weird. I added a breakline to fix it.

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -190,7 +193,7 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem)
                 // we may stop if target is behind queryUntil and we don't have more events to fetch
                 case Some(target) if !hasMoreEvents && target <= queryUntil.maxOrdering => Stop
                 // We may also stop if we have found an event with an offset >= target
-                case Some(target) if xs.exists(_.offset.value >= target) => Stop
+                case Some(target) if xs.exists(_.offset.value >= target)                => Stop
```
renatocaval: Also scalariform. I don't get why I'm getting it now.

src/main/scala/akka/persistence/jdbc/util/SlickDriver.scala
```diff
@@ -17,22 +17,53 @@
...
+          )
+        )
+        case x => throw new SlickException("Expected a DataSource for JNDI name " + name + ", but got " + x)
+      }
+    }
+
+    slickConfiguration.jndiName
+      .map(buildDbFromJndiDatasource)
+      .orElse {
+        slickConfiguration.jndiDbName.map(namingContext.lookup(_).asInstanceOf[Database])
```
renatocaval: Minor optimization. We only do a second JNDI lookup if needed.

src/test/scala/akka/persistence/jdbc/configuration/AkkaPersistenceConfigTest.scala
```diff
@@ -220,6 +220,11 @@ class AkkaPersistenceConfigTest extends FlatSpec with Matchers {
     cfg.slickConfiguration.jndiDbName shouldBe None
     cfg.slickConfiguration.jndiDbName shouldBe None
 
+    cfg.slickConfiguration.asyncExecutorConfig.numThreads shouldBe 20
```
renatocaval: I'm not convinced that this is a useful thing to do. 

The philosophy behind typesafe config is that we should provide default values and let the user override them. This test is only verifying that we do have default values for every property, but that is already provided by resource.conf

src/main/scala/akka/persistence/jdbc/util/SlickDriver.scala
```diff
@@ -17,22 +17,53 @@
...
+        case dataSource: DataSource => Database.forDataSource(
+          ds = dataSource,
+          maxConnections = Option(asyncExecConfig.maxConnections),
+          executor = AsyncExecutor(
+            name = "AsyncExecutor.default",
+            minThreads = asyncExecConfig.minConnections,
+            maxThreads = asyncExecConfig.numThreads,
+            queueSize = asyncExecConfig.queueSize,
+            maxConnections = asyncExecConfig.maxConnections
+          )
```
renatocaval: Just realised that this is wrong. This implementation is doing exactly what I described to be a problematic solution. 
See comment here:
https://github.com/dnvriend/akka-persistence-jdbc/issues/142#issuecomment-345085440

@WellingR suggestion to divide the number of connection across all `AsyncExecutor`s might work, but I believe it's error-prone. Users will have to size`maxConnections` in a non-intuitive fashion.

renatocaval: I'm trying to devise a better solution for that. 

Maybe the simplest solution here is to say. Each component has its own configuration and we do check that they are not sharing JDNI resources. Users may configure three different `DataSource` and have different `AsyncExecutor`s associated and correctly sized for them. 

That means, we just keep the current configs, but we make sure that a `DataSource` is never shared. We let the plugin crash in case of sharing.

To achieve this, we must have something that is guaranteed to run a single time and before loading any of the three components. That bootstrap component will have to check if there is no `DataSource` sharing. Plus, it should read some global config and eventually configure a Slick Database and register it on JNDI.

TimMoore: I don't think it should prohibit sharing data sources. I think this is the primary reason for using JNDI in the first place (e.g., https://github.com/dnvriend/akka-persistence-jdbc/issues/24). Otherwise, you have no elasticity between the connection pools in the three components. You could have a situation where, say the snapshot plugin is sitting on idle connections while the write journal is unable to obtain one.

I acknowledge that this resource concern also exists for threads in the executor. Maybe a good solution would be to ensure that there is exactly one `AsyncExecutor` for each JNDI name (maybe by also storing it in the JNDI context)?

The docs could also be updated to discuss some of these concerns and try to steer people away from `jndiName` in favour of `jndiDbName`.

renatocaval: You are right about sharing JNDI resources. It makes no sense to prohibit because that's exactly the intention.

The problem with have with the current config format is that which piece has it's own `jndiName`, `maxConnections`, `numThreads`, etc. 

So, if we try to create one single `AsyncExecutor`, which properties should we use? 

We need a global config that should be used in case of sharing DS. This config must be read at bootstrap, create a properly configured Slick Database and associate it to `jndiDbName`.  Then, each component willing to use it will have to have a `jndiDbName`, but not a `jndiName`.

WellingR:
FYI, I am working on a change such that the slick configuration is loaded by an akka extenstion. See https://github.com/dnvriend/akka-persistence-jdbc/compare/enable-single-db-configuration

Using this, I want to create the possibility to create a single slick database which is used for the journal, read-journal and for the snapshots. Maybe this resolves the problems above.

renatocaval:
@WellingR yes, that will help. 

We will need a global config thing as well to have one single config for the `AsyncExecutor`.

renatocaval:
closing because superseded by #151

PR #151, opened 2017-11-30
Allow for single database configuration

WellingR:
This PR enables more flexibility in the database configuration.

The goals of this change are the following:

- *Backwards combatibility*: Existing configurations should work as before
- *Flexibility*: It should be possible to provide a custom provider for a database, allowing any possible configuration.
- *Single db for all journals*: It should be easy to configure one single database which is used for all journals.

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
+++ b/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
@@ -22,7 +22,7 @@ import akka.actor.{ ActorSystem, ExtendedActorSystem }
 import akka.persistence.jdbc.config.JournalConfig
 import akka.persistence.jdbc.journal.JdbcAsyncWriteJournal.WriteFinished
 import akka.persistence.jdbc.journal.dao.JournalDao
-import akka.persistence.jdbc.util.{ SlickDatabase, SlickDriver }
+import akka.persistence.jdbc.util.SlickExtension
 import akka.persistence.journal.AsyncWriteJournal
 import akka.persistence.{ AtomicWrite, PersistentRepr }
 import akka.serialization.{ Serialization, SerializationExtension }
@@ -45,11 +45,12 @@ class JdbcAsyncWriteJournal(config: Config) extends AsyncWriteJournal {
   implicit val mat: Materializer = ActorMaterializer()
   val journalConfig = new JournalConfig(config)
 
-  val db: Database = SlickDatabase.forConfig(config, journalConfig.slickConfiguration)
+  val slickExtension = SlickExtension(system)
+  val db: Database = slickExtension.database(config)
 
   val journalDao: JournalDao = {
     val fqcn = journalConfig.pluginConfig.dao
-    val profile: JdbcProfile = SlickDriver.forDriverName(config)
+    val profile: JdbcProfile = slickExtension.profile(config)
     val args = Seq(
       (classOf[Database], db),
       (classOf[JdbcProfile], profile),
@@ -97,7 +98,10 @@ class JdbcAsyncWriteJournal(config: Config) extends AsyncWriteJournal {
       .map(_ => ())
 
   override def postStop(): Unit = {
-    db.close()
+    if (journalConfig.useSharedDb.isEmpty) {
+      // Since a (new) db is created when this actor (re)starts, we must close it when the actor stops
+      db.close()
+    }
     super.postStop()
   }
 
--- a/README.md
+++ b/README.md
@@ -65,14 +65,54 @@ Configure `slick`:
 
 ## Configuration
 
-- [Default](https://github.com/dnvriend/akka-persistence-jdbc/blob/master/src/main/resources/reference.conf)
+akka-persistence-jdbc provides the defaults as part of the [reference.conf](https://github.com/dnvriend/akka-persistence-jdbc/blob/master/src/main/resources/reference.conf)
+this file documents all the values which can be configured.
+
+There are several possible ways to configure loading your database connections. Options will be explained below.
+
+### One database connection pool per journal type
+ 
+There is the possibility to create a separate database connection pool per journal-type (one pool for the write-journal,
+one pool for the snapshot-journal, and one pool for the read-journal). This is the default and the following example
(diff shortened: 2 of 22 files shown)


src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
```diff
@@ -98,7 +99,7 @@ class JdbcAsyncWriteJournal(config: Config) extends AsyncWriteJournal {
       .map(_ => ())
 
   override def postStop(): Unit = {
-    db.close()
+    db.close() // TODO maybe not?
```
renatocaval: I find this tricky. I wasn't aware of that before I faced the issue related with JNDI `DataSource` and `AsyncExecutor`.

The problem that I see with this, and probably also the reason why you added this TODO, is that each plugin will be shutting down at different moments, but if they are sharing the same DB we may be closing it too soon.

I checked how the `akka-persistence-cassandra` is dealing with that and they simply have an independent cassandra session per plugin component.

As of today, the `akka-persistence-jdbc` is not sharing the db, so closing it independently is not an issue. Except that, in Lagom, we share the DB via JNDI. In that case, this is a problem for Lagom, but not for general usage of this plugin.

I think we must have a more robust way of sharing the database (or not) and at the same time, make sure they are all closed in a proper fashion. 

Maybe we should control it through `DefaultStickDatabaseProvider`.  We must somehow make it aware of each DB instance is being used by each plugin (journal, snapshot and read-journal). If they are being shared or not, etc. 

The close method should be delegated to `DefaultStickDatabaseProvider`. So, for example, if it knows that a DB is being shared by the journal and the read-journal, but not by the snapshot store, the first close call will have no effect, while the second will effectively close the shared DB.

WellingR: I think we should make the shutdown strategy configurable. The following configurations would be options:
- Close when the Journal or SnapshotJournal is closed (not possible for the ReadJournal). I agree that in case of database-sharing, we should have a mechanism for closing the db which knows if a database is used elsewhere.
- Close when the ActorSystem is closed (using CoordinatedShutdown from Akka 2.5), in this case we do not need to know if the database is used elsewhere.
- Do not close the db at all (allowing the user to do something manually)

renatocaval: Maybe we should opt for `CoordinatedShutdown` in Akka 2.5. and don't backport it to 2.4 branch.

The current impl in 2.4 is not that bad, because it's not sharing the DB. Unless it's being shared through JNDI.

WellingR:
Note: this is on hold for a while since I currently do not have a lot of time to work on this.

WellingR:
Implementation is now complete: what remains to be done is the following:

- Some of our tests should use a shared-db configuration (we should include example configurations in our tests)
- Our readme should have a small section about these conifguration options

WellingR:
@renatocaval I have updated this PR and it should be ready, can you have another look?

renatocaval:
@WellingR thanks for that. And my apologies for the delay, I was traveling last week.

This PR is certainly an improvement compared to the current situation. I was just wondering if we could make it even more flexible.

As it is now, the slick configuration used for the `jdbc-journal` is the one to be shared. It's possible to configure it as:

```
jdbc-journal.use-shared-db = false
jdbc-snapshot-store.use-shared-db = true
jdbc-read-journal.use-shared-db = true
```

In which case, `jdbc-snapshot-store` and `jdbc-read-journal` will share an instance based on `jdbc-journal.slick`, but `jdbc-journal` will have it's own instance based on the same config.

I was wondering if we could have the following instead:

```
akka-persistence-jdbc {

  database-provider-fqcn = "akka.persistence.jdbc.util.DefaultSlickDatabaseProvider"

  shared-dbs  {
    "db-config-1" = {
      numThreads = 20
      maxConnections = 20
      minConnections = 1
      connectionTimeout = 1800000 // 30 minutes
     }
     
     "db-config-2" = {
       numThreads = 20
       maxConnections = 20
       minConnections = 1
       connectionTimeout = 1800000 // 30 minutes
      }
  }
}
```

and then:

```
// snapshot and query share a DB instance
jdbc-journal.use-shared-db = "db-config-1"
jdbc-snapshot-store.use-shared-db = "db-config-2"
jdbc-read-journal.use-shared-db = "db-config-2"
```
or 

```
// write and query share a DB instance
jdbc-journal.use-shared-db = "db-config-1"
jdbc-snapshot-store.use-shared-db = "db-config-2"
jdbc-read-journal.use-shared-db = "db-config-1"
```

The `SlickExtension` will have to read the `akka-persistence-jdbc` config and configure all the DB it finds under `shared-dbs` and save then in a `Map[String, DB]`.  We could have a wrapper around the DB and have it lazy loaded inside the wrapper object, for instance `Map[String, DBHolder]`. `DBHolder` will have a reference to the config and creates the DB only when requested for the first time.

When `use-shared-db` is used, `{component}.slick` section is ignored and we lookup the map for the passed key.

WDYT?

WellingR:
I like your idea. However one problem I have is that the slick extension needs to be thread safe. Therefore we need a way to safely create the database for the given config. Preferably in a non-blocking but synchronous way.

Do you have any suggestions on how to do this?

renatocaval:
That's a good point. I don't think we are far from it. 

An extension is guaranteed to be loaded only once. `SlickExtensionImpl` is creating a `SlickDatabaseProvider` in its constructor. If we make `SlickDatabaseProvider` to initialise immediately all DB it finds in the config and add it to a immutable `Map` we are safe. Everything will be initalised exactly one by Akka extension mechanism.

The drawback is that we will initialise them eagerly. That's why I suggested the `DBHolder` thing, but it's probably a premature optimisation.

WellingR:
Thanks for this suggestion! I will implement this soon.

WellingR:
@renatocaval I have implemented the changes. (including use of the DbHolder)
Because I wanted to keep the configurations simple I rebased this branch onto the branch of #184 it is probably good to merge that PR first.

I have used the shared db in some of the eventsByTag tests and also in the akka-persistence-tck tests. So the coverage should be good.

PR #152, opened 2017-12-03
Refactor tests

### LINKED ISSUE #149: eventsByTag querying strategy
Hi everyone!

I've noticed that, when trying to query events by tag which is not present in journal from `given offset` the ReadJournal may run quite beefy queries against the database. Load grows while journal size increases, if the events are at the beginning of journal (before `given offset`). After some research I figured out that, when the ReadJournal queries `eventsByTag` with `terminateAfterOffset` = None, empty result set leads to scheduling next query iteration with query condition to start with `given offset` but the previous `max ordering id`. The difference between `given offset` and `max ordering id` continuously increases because the number of events of no interest may grow, which makes further queries more and more heavy (in my case the difference may be up to 1kk). Of course, this happens until event of interest appears at the tail. 

It may look like a bug, what do you think about that?

WellingR:
During my attempt to work on #149 I discovered that many testcases do not re-use the same actor system for all tests within the suite. In many cases this is not harmful. However in our case the JournalSequenceActor is keeping track of some state, which can only be reset by restarting the actor system.

The base testSpecs that we have now are  `SharedActorSystemTestSpec` (which is more or less what we had), and I have added `SingleActorSystemPerTestSpec` in which we can call `withActorSystem` to start an ActorSystem, in addition the database is also restarted for each test.

I have restarted the build several times on travis and it seems stable.
Unfortunately I seem to have some trouble with the oracle testcases locally every now and then...

*Edit:*
It seems that all tests are stable when I run them using sbt (or test.sh). Travis also does not seem to have any build issues


WellingR:
There is nothing that I can do about the codacy issues. Anyhow, these changes are in the tests only.

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -93,7 +93,8 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem)
 
   // Started lazily to prevent the actor for querying the db if no eventsByTag queries are used
   private[query] lazy val journalSequenceActor = system.actorOf(
-    JournalSequenceActor.props(readJournalDao, readJournalConfig.journalSequenceRetrievalConfiguration), "akka-persistence-jdbc-journal-sequence-actor")
+    JournalSequenceActor.props(readJournalDao, readJournalConfig.journalSequenceRetrievalConfiguration), "akka-persistence-jdbc-journal-sequence-actor"
+  )
```
renatocaval: Scalariform is forcing an annoying formatting here by putting the ) on the next line. I have had it as well on the PR I have openned. 

Maybe the best here is to format it as:
```scala
private[query] lazy val journalSequenceActor = 
  system.actorOf(
    JournalSequenceActor.props(readJournalDao, readJournalConfig.journalSequenceRetrievalConfiguration),
    "akka-persistence-jdbc-journal-sequence-actor"
  )
```

As such, we can at least have this trailing ) on the same column as `system.actor` which where the open ( comes from.

src/test/scala/akka/persistence/jdbc/query/QueryTestSpec.scala
```diff
@@ -48,14 +46,11 @@ trait ReadJournalOperations {
   def countJournal: Future[Long]
 }
 
-trait ScalaJdbcReadJournalOperations extends ReadJournalOperations {
-  implicit def system: ActorSystem
+class ScalaJdbcReadJournalOperations(readJournal: JdbcReadJournal)(implicit system: ActorSystem, mat: Materializer) extends ReadJournalOperations {
```
renatocaval: Good idea to use composition instead of inheritance for this. 👍

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
WellingR: Good point

PR #153, opened 2017-12-04
Efficiency improvement in eventsByTag

### LINKED ISSUE #149: eventsByTag querying strategy

WellingR:
#152 has been merged, I have re-targeted this branch to master

fixes #149

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
+++ b/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
@@ -199,7 +199,15 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem) 
                   if (hasMoreEvents) Continue else ContinueDelayed
               }
 
-            val nextStartingOffset = if (xs.isEmpty) from else xs.map(_.offset.value).max
+            val nextStartingOffset = if (xs.isEmpty) {
+              /* If no events matched the tag between `from` and `maxOrdering` then there is no need to execute the exact
+               * same query again. We can continue querying from `maxOrdering`, which will save some load on the db.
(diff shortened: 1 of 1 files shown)


src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -198,7 +198,7 @@ class JdbcReadJournal(config: Config)(implicit val system: ExtendedActorSystem)
                   if (hasMoreEvents) Continue else ContinueDelayed
               }
 
-            val nextStartingOffset = if (xs.isEmpty) from else xs.map(_.offset.value).max
+            val nextStartingOffset = if (xs.isEmpty) math.max(from, queryUntil.maxOrdering) else xs.map(_.offset.value).max
```
renatocaval: maybe we should add a comment to explain why this is needed and which special case we are covering. 

I remember vaguely that we have discussed it, but it's now somewhere deep in my memory.

WellingR:
Added some comments in the code to clarify why this is needed

PR #154, opened 2017-12-05
Added scaladoc to the eventsByTag queries

### LINKED ISSUE #82: eventsByTag returns events for wrong tag
Hi everyone,

When looking for events with tag "User" with `currentEventsByTag`, this also returns events with tag "UserEmail". In my opinion only events with tag "User" should be returned. I currently experience the problem when using the default dao configuration and a postgres backend.

When searching for the cause of this problem I found that in `package akka.persistence.jdbc.dao.bytea.readjournal` in [`trait BaseByteArrayReadJournalDao`](https://github.com/dnvriend/akka-persistence-jdbc/blob/master/src/main/scala/akka/persistence/jdbc/dao/bytea/readjournal/ByteArrayReadJournalDao.scala) events are filtered by `s"%$tag%"`. Might this be a bug?

With kind regards,
Felix

WellingR:
Added documentation for the eventsByTag an currentEventsByTag queries.

Specifically the documentation describes the meaning of the `offset` in our plugin, it states that the streams are ordered by `offset`. And that the query is implemented using a LIKE query.

Adding this documentation fixes #82. Because this information is needed for users to be able to work around the tag substring issue as described in #82. (Now that we have documented this, we can consider this a feature instead of a bug 😉)

The actual content of the docs has been based on what is originally described in the scaladoc of the akka traits for [CurrentEventsByTagQuery](https://github.com/akka/akka/blob/master/akka-persistence-query/src/main/scala/akka/persistence/query/scaladsl/CurrentEventsByTagQuery.scala) and [EventsByTagQuery](https://github.com/akka/akka/blob/master/akka-persistence-query/src/main/scala/akka/persistence/query/scaladsl/EventsByTagQuery.scala).

The other persistence-queries do not need any additional documentation, since the documentation provided by the (inherited) akka scaladoc is already sufficient.

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/query/javadsl/JdbcReadJournal.scala
+++ b/src/main/scala/akka/persistence/jdbc/query/javadsl/JdbcReadJournal.scala
@@ -46,9 +46,41 @@ class JdbcReadJournal(journal: ScalaJdbcReadJournal) extends ReadJournal
   override def eventsByPersistenceId(persistenceId: String, fromSequenceNr: Long, toSequenceNr: Long): Source[EventEnvelope, NotUsed] =
     journal.eventsByPersistenceId(persistenceId, fromSequenceNr, toSequenceNr).asJava
 
+  /**
+    * Same type of query as [[EventsByTagQuery#eventsByTag]] but the event stream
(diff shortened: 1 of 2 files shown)


renatocaval:
Thanks, for making it a feature! 😄

PR #158, opened 2017-12-12
Journal sequence actor uses NumericRanges instead of a Set to keep track of missing elements

### LINKED ISSUE #155: JournalSequenceActor can be very slow and even cause out-of-memory errors in case of huge sequence gaps
Given the following case:

The journal has one entry, where the ordering column has value 29721004 (almost 30 million)
(appearently the ordering_seq did not start at 1 when i saw this issue occur.

When an eventsByTag query is started, the journal sequence actor will load. And start querying, unfortunately the code was not build with the expectations that "gaps" could be this large. And it starts creating sequences of missing ordering ids very inefficiently. This cases the application to go out of memory (when 500MB memory is allocated to the JVM).

We should re-analyse the design of the journal sequence actor so that it is tolerant for cases like these.

WellingR:
This PR fixes #155

Before this fix, the journal sequence actor would build several scala `Set`s containing all ordering ids which were considered missing. In case that extreme amounts of events were missing (which can be normal if the first event does not start at 1, but instead of a very high value (millions). Then the set construction would be very inefficiënt and memory-consuming.

The solution is a wrapper type `MissingElements` which internally contains a `Seq[NumericRange[OrderingId]]`. A numericRange uses a constant amount of memory regardless of its size. Which is exactly what we need in this case.

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
+++ b/src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
@@ -1,4 +1,5 @@
-package akka.persistence.jdbc.query
+package akka.persistence.jdbc
+package query
 
 import akka.actor.{Actor, ActorLogging, Props, Status, Timers}
 import akka.persistence.jdbc.query.dao.ReadJournalDao
@@ -7,6 +8,7 @@ import akka.persistence.jdbc.config.JournalSequenceRetrievalConfig
 import akka.stream.Materializer
 import akka.stream.scaladsl.Sink
 
+import scala.collection.immutable.NumericRange
 import scala.concurrent.duration.FiniteDuration
 
 object JournalSequenceActor {
@@ -25,6 +27,26 @@ object JournalSequenceActor {
   private case object AssumeMaxOrderingIdTimerKey
 
   private type OrderingId = Long
+
+  /**
+    * Efficient representation of missing elements using NumericRanges.
+    * It can be seen as a collection of OrderingIds
+    */
+  private case class MissingElements(elements: Seq[NumericRange[OrderingId]]) {
+    def ++(range: NumericRange[OrderingId]) = MissingElements(elements :+ range)
+    def contains(id: OrderingId): Boolean = elements.exists(_.containsTyped(id))
+    def isEmpty: Boolean = elements.forall(_.isEmpty)
+
+    /**
+      * Preserve only the NumericRanges where the end (last value) is larger than the given id.
+      */
+    def filterLargerThan(id: OrderingId): MissingElements = {
+      MissingElements(elements.filter(_.end > id))
+    }
+  }
+  private object MissingElements {
+    def empty: MissingElements = MissingElements(Vector.empty)
+  }
 }
 
 /**
@@ -56,7 +78,7 @@ class JournalSequenceActor(readJournalDao: ReadJournalDao, config: JournalSequen
    * @param moduloCounter A counter which is incremented every time a new query have been executed, modulo `maxTries`
    * @param previousDelay The last used delay (may change in case failures occur)
    */
-  def receive(currentMaxOrdering: OrderingId, missingByCounter: Map[Int, Set[OrderingId]], moduloCounter: Int, previousDelay: FiniteDuration = queryDelay): Receive = {
+  def receive(currentMaxOrdering: OrderingId, missingByCounter: Map[Int, MissingElements], moduloCounter: Int, previousDelay: FiniteDuration = queryDelay): Receive = {
 
     case ScheduleAssumeMaxOrderingId(max) =>
       // All elements smaller than max can be assumed missing after this delay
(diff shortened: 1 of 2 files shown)


src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
```diff
@@ -139,15 +152,10 @@ class JournalSequenceActor(readJournalDao: ReadJournalDao, config: JournalSequen
...
-      (missingByCounter + (moduloCounter -> missingElems))
-        .map {
-          case (key, value) =>
-            key -> value.filter(missingId => missingId > nextMax)
-        }
+    val newMissingByCounter = missingByCounter + (moduloCounter -> missingElems)
 
     // did we detect gaps in the current batch?
-    val noGapsFound = newMissingByCounter.values.forall(_.isEmpty)
+    val noGapsFound = missingElems.isEmpty
```
WellingR: Since we no longer filter elements which were already missing in previous iterations. I was able to make this much more simple

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
```diff
@@ -117,14 +132,12 @@ class JournalSequenceActor(readJournalDao: ReadJournalDao, config: JournalSequen
 
             // if it's a gap and has been detected before on a previous iteration we give up
             // that means that we consider it a genuine gap that will never be filled
-            case e if givenUp(e)               => missing
```
renatocaval: IIUC, there was a bug here. Because `e` was never added to the missing Seq, this check would never return true and we would always add the gap in the next case (if not already missing).

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
```diff
@@ -117,14 +132,12 @@ class JournalSequenceActor(readJournalDao: ReadJournalDao, config: JournalSequen
...
+            case e if (previousElement + 1 until e).forall(givenUp.contains) => missing
 
             // any other case is a gap that we expect to be filled soon
             case _ =>
               val currentlyMissing = previousElement + 1 until currentElement
-              // we don't want to declare it as missing if it has been already declared on a previous iterations
-              def alreadyMissing(e: Long) = missingByCounter.values.exists(_.contains(e))
-              missing ++ currentlyMissing.filterNot(alreadyMissing)
+              missing ++ currentlyMissing
```
renatocaval: Would it make more sense if we let the `++` method decide if a missing range is already recorded or not? As such, we could remove the case on line 135. 

Basically, the code has two possible branches: 
1. there is no gap and it continues
2. there is a gap and we record it. 

As such, we can even replace the pattern matching by a simple `if/else`. 

```scala
val newMissing =
  if(previousElement + 1 == currentElement) 
    missing
  }
  else {
    val currentlyMissing = previousElement + 1 until currentElement
    missing ++ currentlyMissing
  }
```

or this:

```scala
val newMissing =
  if(previousElement + 1 == currentElement) missing
  else missing.addRange(previousElement, currentElement)
```

`addRange` would be reponsible to check if it makes sense to add the range or not.

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
```diff
@@ -25,6 +27,19 @@ object JournalSequenceActor {
   private case object AssumeMaxOrderingIdTimerKey
 
   private type OrderingId = Long
+
+  /**
+    * Efficient representation of missing elements using NumericRanges.
+    * It can be seen as a collection of OrderingIds
+    */
+  private case class MissingElements(elements: Seq[NumericRange[OrderingId]]) {
```
renatocaval: I like this! 
I'm always in favour of micro-models to increase clarity! 👍

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
```diff
@@ -136,18 +136,18 @@ class JournalSequenceActor(readJournalDao: ReadJournalDao, config: JournalSequen
               currentElement
             } else currentMax
 
+          // we accumulate in newMissing the gaps we detect on each iteration
+          val newMissing =
+            if (previousElement + 1 == currentElement || newMax == currentElement) missing
+            else missing.addRange(previousElement + 1, currentElement)
```
WellingR: Based on the previous PR comment I managed to simplify this piece of code. I did have to add the `|| newMax == currentElement` condition though (otherwise some elements which were part of givenUp were included in the missing elements collection).

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
```diff
@@ -25,6 +27,22 @@ object JournalSequenceActor {
...
   private type OrderingId = Long
+
+  /**
+    * Efficient representation of missing elements using NumericRanges.
+    * It can be seen as a collection of OrderingIds
+    */
+  private case class MissingElements(elements: Seq[NumericRange[OrderingId]]) {
+    def addRange(from: OrderingId, until: OrderingId): MissingElements = {
+      val newRange = from.until(until)
+      MissingElements(elements :+ newRange)
```
renatocaval: Don't we need to check that `newRange` is not already added? Previously, we were writing:

```scala
case e if (previousElement + 1 until e).forall(givenUp.contains) => missing
```
So, when the range was already there, we were just skipping it. The `addRange` could do the same, no?

renatocaval: Well, I'm reconsidering that. Actually, I think it's impossible to produce the same range or a sub-range, so such a check will always return false we the range will be added anyway.

WellingR: Previously were were writing `givenUp.contains` however the MissingElements which we are accumulating are not the missing elements which are `givenUp`.

This has actually already been handled in the code of the actor:
```
           // we must decide if we move the cursor forward
            val newMax =
              if ((currentMax + 1).until(currentElement).forall(givenUp.contains)) {
               // we move the cursor forward when:
               // 1) they have been detected as missing on previous iteration, it's time now to give up
               // 2) current + 1 == currentElement (meaning no gap). Note that `forall` on an empty range always returns true
                currentElement
              } else currentMax
  
           // we accumulate in newMissing the gaps we detect on each iteration
           val newMissing =
             if (previousElement + 1 == currentElement || newMax == currentElement) missing
             else missing.addRange(previousElement + 1, currentElement)
```

We compute newMax first which contains the check `((currentMax + 1).until(currentElement).forall(givenUp.contains))`

Because of this we can re-use the information (the value of newMax) in the condition for newMissing: ` if (previousElement + 1 == currentElement || newMax == currentElement) missing`

renatocaval: Fair enough. 

I will merge this one then.

PR #161, opened 2018-01-10
Implement optional physical deletion of marked journal messages

### LINKED ISSUE #160: Add option to physically delete journal messages
Currently journal items are marked as deleted when `akka.persistence.Eventsourced#deleteMessages` is called. It happens inside `akka.persistence.jdbc.journal.dao.BaseByteArrayJournalDao#delete` method.

When it comes to bigger amounts of journal items being written and deleted (after snapshot, e.g.) this functionality may bring operational issue resulting in lack of disk space on DB server.

What do you think about introducing a configuration setting that will instruct `JournalDao` to physically delete journal items from DB, not just mark them as deleted.

For instance `cassandra` journal implementation does permanent deletion by default [see here](https://github.com/krasserm/akka-persistence-cassandra/blob/master/src/main/scala/akka/persistence/cassandra/journal/CassandraStatements.scala)

@dnvriend do you assume any technical or architectural drawback in this proposal ?

dmi3zkm:
Fixes #160

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -44,6 +44,8 @@ jdbc-journal {
   batchSize = 400
   # The maximum number of batch-inserts that may be running concurrently
   parallelism = 8
+  # Only mark as deleted. If false, delete physically
+  logicalDelete = true
 
   slick {
     
--- a/src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalDao.scala
+++ b/src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalDao.scala
@@ -22,10 +22,10 @@ import akka.persistence.jdbc.config.JournalConfig
 import akka.persistence.jdbc.serialization.FlowPersistentReprSerializer
 import akka.persistence.{AtomicWrite, PersistentRepr}
 import akka.serialization.Serialization
-import akka.stream.{Materializer, OverflowStrategy, QueueOfferResult}
 import akka.stream.scaladsl.{Keep, Sink, Source}
-import slick.jdbc.JdbcProfile
+import akka.stream.{Materializer, OverflowStrategy, QueueOfferResult}
 import slick.jdbc.JdbcBackend._
+import slick.jdbc.JdbcProfile
 
 import scala.collection.immutable._
 import scala.concurrent.{ExecutionContext, Future, Promise}
@@ -44,8 +44,8 @@ trait BaseByteArrayJournalDao extends JournalDao {
   implicit val ec: ExecutionContext
   implicit val mat: Materializer
 
+  import journalConfig.daoConfig.{batchSize, bufferSize, logicalDelete, parallelism}
   import profile.api._
-  import journalConfig.daoConfig.{batchSize, bufferSize, parallelism}
 
   private val writeQueue = Source.queue[(Promise[Unit], Seq[JournalRow])](bufferSize, OverflowStrategy.dropNew)
     .batchWeighted[(Seq[Promise[Unit]], Seq[JournalRow])](batchSize, _._2.size, tup => Vector(tup._1) -> tup._2) {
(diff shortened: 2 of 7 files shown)


src/main/resources/reference.conf
```diff
@@ -44,6 +44,8 @@ jdbc-journal {
   batchSize = 400
   # The maximum number of batch-inserts that may be running concurrently
   parallelism = 8
+  # Only mark as deleted. If false, delete physically
+  markDelete = true
```
renatocaval: Can we call it `logicalDelete`?

src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalDao.scala
```diff
@@ -92,9 +92,20 @@ trait BaseByteArrayJournalDao extends JournalDao {
...
-  } yield ()
+  override def delete(persistenceId: String, maxSequenceNr: Long): Future[Unit] =
+    if (markDelete) {
+      for { _ <- db.run(queries.markJournalMessagesAsDeleted(persistenceId, maxSequenceNr)) } yield ()
+    } else {
+      for {
+        _ <- db.run(queries.markJournalMessagesAsDeleted(persistenceId, maxSequenceNr))
+        highestMarkedSequenceNr <- highestMarkedSequenceNr(persistenceId)
+        _ <- db.run(queries.delete(persistenceId, highestMarkedSequenceNr - 1))
+      } yield ()
```
renatocaval: I don't get the intention here.
The first call is marking them all as logically delete, then we take the highest number and do a hard delete from everything below it. Why?
Why can't we simply delete them all without first marking them? For instance:

```scala
val result = 
  if (markDelete) {
    db.run(queries.markJournalMessagesAsDeleted(persistenceId, maxSequenceNr))
  } else {
    db.run(queries.delete(persistenceId, highestMarkedSequenceNr - 1))
  }
  result.map(_ => ())
```
(btw, we don't need for comprehension here)

There is another issue with this code. Each call to `db.run` is an apart transaction. Although we know that nobody can change past events we may want to run it (and hit the DB) in a single transaction. 

But again, I don't think we should first mark them as deleted. Let's just delete them.

dmi3zkm: There is a test named `A journal must not reset highestSequenceNr after journal cleanup` in `JournalSpec` in `akka-persistence-tck` which won't pass if messages simply deleted to highest sequence number. That's why messages are marked and then deleted up to `(highestSequenceNr - 1)`.

src/main/resources/reference.conf
dmi3zkm: yes, sure

src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalDao.scala
renatocaval: Interesting and strange! 

We need to think what to do that and how this will affect the query side.

For instance, if we have a journal with the following evetns:

```
E1
E2
E3
E4
E5
```
and we do a hard delete up to offset 5, we will end up with:

```
E1 (hard deleted)
E2 (hard deleted)
E3 (hard deleted)
E4 (hard deleted)
E5 (soft deleted)
```

And then comes the quesiton, what to do on the query side? I would say, a soft deleted show never be delivered. Not when replaying an actor (re-hydratation) and nor when querying (see #104). 

I need to check the internals and reason for that, but that makes me think that the latest event can't be hard deleted because we need its sequence nr to build the `RecoverySuccess` message.


The testkit code says...

```scala 
 "not reset highestSequenceNr after journal cleanup" in {
      journal ! ReplayMessages(0, Long.MaxValue, Long.MaxValue, pid, receiverProbe.ref)
      1 to 5 foreach { i ⇒ receiverProbe.expectMsg(replayedMessage(i)) }
      receiverProbe.expectMsg(RecoverySuccess(highestSequenceNr = 5L))

      journal ! DeleteMessagesTo(pid, Long.MaxValue, receiverProbe.ref)
      receiverProbe.expectMsg(DeleteMessagesSuccess(Long.MaxValue))

      journal ! ReplayMessages(0, Long.MaxValue, Long.MaxValue, pid, receiverProbe.ref)
      receiverProbe.expectMsg(RecoverySuccess(highestSequenceNr = 5L))
}
```
As we can see, the messages are not replayed, but we need to keep the knowledge about the `highestSequenceNr`.

The conclusion is, a journal implementing hard delete, must also support soft delete because of `highestSequenceNr`. And it's query should not include the soft deleted event.

renatocaval: @dmi3zkm in any case, I think that the code should be:

```scala

val actions = 
  for {
    _ <- queries.markJournalMessagesAsDeleted(persistenceId, maxSequenceNr)
   highestMarkedSequenceNr <- highestMarkedSequenceNr(persistenceId)
   _ <- queries.delete(persistenceId, highestMarkedSequenceNr - 1)
  } yield ()

db.run(actions.transactionaly)
```

Note, `highestMarkedSequenceNr` must return a `DBIOAction` instead of `Future`.

So we hit the DB only once. 

And it will also be good to add a comment explaining why we are leaving the latest event as a soft delete.

patriknw: It's right that we are not allowed to "reuse" sequence numbers. If E1-E5 are deleted the next written event must be E6 (seqnr=6). Exactly how to implement that is up to the journal implementation.

renatocaval:
Thanks @dmi3zkm. 

I think we can merge it, but I would link it to #104 before we make a new release.

@WellingR, WDYT?

PR #162, opened 2018-01-20
[WIP] added support for filtering out deleted events

renatocaval:
Originally from #104 initiated by @morozov1.

This PR introduces filtering of deleted events. To better understand why this is needed, we must understand the historical reasons for this and make some guesses about its evolution.

What follows is an assumption, I don't have all the knowledge of how the plugin architecture evolved. Therefore will try to outline here all the information I could collect and the assumptions I'm making.

### Delete Journal Events
I guess that everybody agrees that an Event Sourcing application should not delete events. However, sometimes this may be needed (free up space) or required by law (currently flawed because of snapshoting and highest seq num issue, read further). So, this feature makes sense but has to be used with caution. 

### Physical or Logical Delete
Events can be physically or logically deleted. Logical deletion seems at first a very odd feature because it doesn't help at all in freeing up space nor does it fulfill legal regulations.

As far I can understand, the only reasons for logical deletions are:

1. quarantine - events are deleted and after some period, and if nothing gets broken, events are effectively deleted by a DB admin. 
2. plugin spec requires that the highest sequence number ever used by a persistent actor is not lost. How this is achieved is an implementation detail. In the case of akka-persistence-jdbc, this is done by performing a logical delete on the latest event. The event is kept because of the highest seq num requirement, but it's invisible to the outside.

### Logical Deletes and Backward "awkward" Compatibility
The previous behavior was *logical deletion* with events being delivered on query side. This is very odd. Why one would perform a logical delete on the write-side and still consume the events on the read-side? If the events are still needed, one should not "delete" them. 

This PR preserves this behavior. Logical deletes are enable by default and events are delivered when querying.

### Physical Deletes
Physical deletes are now possible (see #161, thanks @dmi3zkm). However, as mentioned above, the last event is not physically delete because we need to be able to trace the highest seq num per persistent actor. Because of the logical deletion of latest event is an internal trick, we should treat it as a real delete, therefore they are not delivered on queries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,13 +238,13 @@ class TaggingEventAdapter extends WriteEventAdapter {
   def withTag(event: Any, tag: String) = Tagged(event, Set(tag))
 
   override def toJournal(event: Any): Any = event match {
-    case _: PersonCreated ⇒
+    case _: PersonCreated =>
       withTag(event, "person-created")
-    case _: FirstNameChanged ⇒
+    case _: FirstNameChanged =>
       withTag(event, "first-name-changed")
-    case _: LastNameChanged ⇒
+    case _: LastNameChanged =>
       withTag(event, "last-name-changed")
-    case _ ⇒ event
+    case _ => event
   }
 }
 ```
--- a/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
+++ b/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
@@ -18,12 +18,13 @@ package akka.persistence.jdbc.journal
 
 import java.util.{ HashMap => JHMap, Map => JMap }
 
+import akka.Done
 import akka.actor.{ ActorSystem, ExtendedActorSystem }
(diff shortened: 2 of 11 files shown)


renatocaval:
We need extra test for event delete. 

We must prove that:
- [ ] logical deleted events are still delivered by queries (previous behavior)
- [ ] lastest event that is supposed to be deleted, but is kept as logically deleted, is NOT delivered on queries.

renatocaval:
@WellingR yep, iti was scalariform. At least for the the scaladoc.

I'm up for re-evaluating the formatting options. See my comment here:
https://github.com/dnvriend/akka-persistence-jdbc/pull/163#issuecomment-359362014

renatocaval:
@WellingR tests covering the hard deletion. Hard deleted events are not included in queries, logically deleted events are included.

I need to check the override of config property. Got some issues with the tests. I will check it tomorrow.

renatocaval:
I don't know what happened here. PR closed, but I didn't do it. And I can't reopen it. I think it got automatically closed because I rewrote the history on my own repo. I will send another one.

PR #163, opened 2018-01-20
upgrade to sbt-scalariform 1.8.2

sullis:


WellingR:
Thanks for your contribution.

I am actually not a fan of scalariform at all. Because I keep on ending up revering changes made by the plugin to prevent otherwise unchanged files form being reformatted. I am actually more in favour of disabling this plugin as a while.

@renatocaval what do you think?

renatocaval:
@WellingR I also don't like scalariform for the reason you mentioned as well. 

The major problem with scalariform is that the sbt plugin and IntelliJ plugin doesn't share information.

scalafmt is not only a better option for formatting (IMO), but the configuration is based on a simple test file that is read by all existing plugins (sbt, intellij, etc).

Whenever I work on a project with scalafmt, I configure my ide to format on save and as a result, I don't get disturbed by sbt reformatting the file in the background. 

I'm ok to merge this one for the moment and send another PR with scalariform replaced by scalafmt

WellingR:
Eventually replacing scalariform with scalafmt would be a good idea. For now I will merge this

WellingR:
Thanks @sullis

PR #174, opened 2018-03-20
make the use of sys.addShutdownHook(db.close()) configurable.

### LINKED ISSUE #171: Database closing outside of Coordinated Shutdown
When my application down with SIGTERM, the database is closed immediately causing some errors.

To prevent this problem and have more control, I ask you if is possible to close the database inside of Coordinated Shutdown changing this code `sys.addShutdownHook(db.close())` in `akka.persistence.jdbc.query.JdbcReadJournal` line 72, to be inside of some phase of Coordinated Shutdown, maybe on “before-actor-system-terminate” or “actor-system-terminate”

Akka version: 2.5.9

notNotDaniel:
We'd like to call db.close() as part of our own shutdown process, and having it called automatically by a shutdown hook is problematic. This change adds a new config option to disable adding the hook automatically (with the default being to preserve the current behavior).

fixes #171


WellingR:
Thanks for this!

agnaldo4j:
Thanks!!

PR #180, opened 2018-04-10
Added support for serializing using Akka serialization

jroper:
Currently, this plugin serializes events by using Java serialization of `akka.persistence.PersistentImpl`. Given that `akka.persistence.PersistentImpl` is a private class, not meant to be exposed to the outside world, this is really bad. Akka could change the structure of this class whenever they want in a future release, and this will render every serialized event impossible to deserialize.

It also means that if someone has registered a custom serializer using Akka serialization for their events, that that custom serializer won't be used, which is in contrast to most if not all other Akka persistence plugins.

This modifies akka-persistence-jdbc to use Akka serialization properly, looking up a serializer for the user supplied event/snapshot, and persisting the id and manifest of that serializer along with the serialized bytes. This means users can register custom serializers for their events using the standard Akka serialization mechanism, and using no custom akka-persistence-jdbc code.

The change is both backwards and forwards compatible. It achieves this by introducing a new column for the serialized data, rather than reusing the existing one. Forwards compatibility is achieved by writing both to the old and the new column, but this has a significant overhead, so it's disabled by default. Backwards compatibility is achieved by reading from the new column if it's not null, otherwise, reading from the old column.

The schemas have all been updated such that the old column is now nullable, and all the new columns are not null. This is suitable for a new project. For an existing project, schema migrations have been provided, that set both the old column and the new columns to nullable, for backwards compatibility.

Finally, an option has been provided that will allow the old column to be deleted altogether, but it is disabled by default.


src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalSerializer.scala
```diff
@@ -17,28 +17,69 @@
...
+        Some(serializer.identifier),
+        Some(serManifest),
+        Some(persistentRepr.writerUuid))
+    }
   }
 
   override def deserialize(journalRow: JournalRow): Try[(PersistentRepr, Set[String], Long)] = {
-    serialization.deserialize(journalRow.message, classOf[PersistentRepr])
-      .map((_, decodeTags(journalRow.tags, separator), journalRow.ordering))
+    if (journalRow.event.isDefined && journalRow.serId.isDefined) {
```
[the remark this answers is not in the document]
jroper: Case matching on a large case class (JournalRow has 11 fields) would be far less readable than this, and in this case we have 3 options that we want to match on, which would require allocating a 3-tuple just to do a simple if, and I think would reduce the readability too and also be more error prone because two of those options have the same type.

jroper:
PR is passing. Please review carefully, especially the schema migrations.

Also, is defaulting forwards compatibility to false a good idea? Also, the schemas still include the `message` column even though it will never be used in a newly created application. Is that ok? It could be removed, if we default `hasMessageColumn` to `false` in `reference.conf`, but this would mean when upgrading, even if not doing rolling upgrades, you would need to set it to `true` to support reading of old events.

WellingR:
Thanks! I had noticed before that the PersistentRepr was serialized as a whole, but I never considered actually changing this because of the extra work needed to keep everything backwards compatible.

I took a look and the implementation looks good, but I need some time to dive into all the details.

About the default settings in the reference.conf:
The general rule is that we want the project to work as good as possible out of the box, this should hold for both existing and for new applications. However we should assume that existing users read a migration guide when they update a major version. (or at least, we should assume that they start reading the guide as soon as they run into trouble).

Following my reasoning above, this means that we can set `hasMessageColumn` to `false` and the same for the compatibility flag

README.md
```diff
@@ -30,6 +30,14 @@ The release notes can be found [here](https://github.com/dnvriend/akka-persisten
 
 For change log prior to v3.2.0, visit [Version History Page (wiki)](https://github.com/dnvriend/akka-persistence-jdbc/wiki/Version-History).
 
+To migrate from v3.x to v4.x, the database schema needs to be updated. To migrate, see the `<database-name>-migration-v4.sql` files in your corresponding databases [schema directory](https://github.com/dnvriend/akka-persistence-jdbc/tree/master/src/test/resources/schema/). The schema update is both backwards and forwards compatible. If you wish to perform rolling upgrades, canary upgrades, or want to be able to rollback, you will need to enable forwards compatibility, which can be done by setting the following configuration:
+
+```
+akka-persistence-jdbc.migration.v4.forwards-compatibility = true
```
WellingR: I am always confused about this. Shouldn't this be backwards compatibility, since this means the new data is still backwards compatible with akka-persistence-jdbc v3?

jroper: It depends on both the context and the direction that you're starting from when you're describing the compatibility. In this case, the context is the compatibility between the code and the data, and I'm starting from the direction of the code (since this is a flag that controls the code, it's not a migration being applied to the data). If the new code is compatible with the old data, then that's backwards compatible. If the old code is compatible with the new data, then that's forwards compatible. So this flag ensures that the new code is writing new data that the old code will be forwards compatible with.

jroper:
I've made the following changes:

* Introduced a `backwards-compatibility` option (in addition to `forwards-compatibility`), which defaults to false, and is used as a default for the `hasMessageColumn`/`hasSnapshotColumn` config.
* Removed the `message`/`snapshot` columns from the schemas, so new applications that create the schema from scratch won't have or try to use these columns.
* Introduced a migration task that can be run in the background (shouldn't require an outage) to migrate old events to store them in the new format - thus eliminating the danger of Akka persistence changing the PersistentImpl class.
* Added tests that load an H2 database that was created using v3 with the old schema, and then tests that things work in backwards compatibility mode, as well as tests that after running the new migration task, things work without backwards compatibility mode on.
* Updated the README to document the backwards compatibility config, as well as the migration task.

src/test/scala/akka/persistence/jdbc/migration/V4MigrationSpec.scala
```diff
@@ -0,0 +1,101 @@
...
+import scala.concurrent.duration._
+
+class V4MigrationSpec extends SimpleSpec with BeforeAndAfter with ClasspathResources {
+
+  var system: ActorSystem = _
+  var tmpDirectory: File = _
+
+  before {
+    tmpDirectory = Files.createTempDirectory("akka-persistence-jdbc-migration-").toFile
+    IOUtils.copyAndClose(fromClasspathAsStream("v3-h2-data.mv.db"), new FileOutputStream(new File(tmpDirectory, "database.mv.db")))
```
WellingR: It is probably a good idea to include this test for each supported database.

WellingR: Use `Files.newOutputStream()` see https://www.cloudbees.com/blog/fileinputstream-fileoutputstream-considered-harmful

src/main/scala/akka/persistence/jdbc/migraition/V4Migration.scala
```diff
@@ -0,0 +1,175 @@
...
+   */
+  private val RowsPerTransaction = 100
+
+  def run(): Unit = {
+    val db = SlickDatabase.forConfig(config, journalConfig.slickConfiguration)
+    try {
+      println(s"Migrating journal events, each . indicates $RowsPerTransaction rows migrated.")
+      val migration = migrateNextBatch(db, 0)
+      val migrated = Await.result(migration, Duration.Inf)
+      println(s"Journal migration complete! $migrated events were migrated.")
```
WellingR: Instead of using println it is probably better to use a logger.

WellingR:
One question I still have

> Currently, this plugin serializes events by using Java serialization of akka.persistence.PersistentImpl.

Are you sure? In one of my project I have configured `akka.actor.allow-java-serialization = off` to make sure that misconfigured serializers are detected. This does not cause any errors. Can you explain why this words?

> It also means that if someone has registered a custom serializer using Akka serialization for their events, that that custom serializer won't be used, which is in contrast to most if not all other Akka persistence plugins.

I always throught this actually worked. However I too was annoyed that the message was wrapped in the PersistentRepr, because I could not actually extract and inspect the serialized data in the database.

src/main/scala/akka/persistence/jdbc/migraition/V4Migration.scala
```diff
@@ -0,0 +1,175 @@
...
+      case batch =>
+        print(".")
+        migrateNextBatch(db, migrated + batch)
+    }
+  }
+
+  private def migrateJournalBatch(db: Database): Future[Int] = {
+    val batchUpdate = JournalTable
+      .filter(_.event.isEmpty)
+      .take(RowsPerTransaction)
```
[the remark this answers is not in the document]
jroper: Cool, I'll update.

src/main/scala/akka/persistence/jdbc/journal/dao/JournalTables.scala
```diff
@@ -27,14 +27,32 @@ trait JournalTables {
...
     val ordering: Rep[Long] = column[Long](journalTableCfg.columnNames.ordering, O.AutoInc)
     val persistenceId: Rep[String] = column[String](journalTableCfg.columnNames.persistenceId, O.Length(255, varying = true))
     val sequenceNumber: Rep[Long] = column[Long](journalTableCfg.columnNames.sequenceNumber)
     val deleted: Rep[Boolean] = column[Boolean](journalTableCfg.columnNames.deleted, O.Default(false))
     val tags: Rep[Option[String]] = column[Option[String]](journalTableCfg.columnNames.tags, O.Length(255, varying = true))
-    val message: Rep[Array[Byte]] = column[Array[Byte]](journalTableCfg.columnNames.message)
+    @deprecated("This was used to store the the PersistentRepr serialized as is, but no longer.", "4.0.0")
+    private val message: Rep[Option[Array[Byte]]] = column[Array[Byte]](journalTableCfg.columnNames.message)
+
+    val event: Rep[Option[Array[Byte]]] = column[Option[Array[Byte]]](journalTableCfg.columnNames.event)
```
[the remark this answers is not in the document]
jroper: As I understand it, the message column isn't included, because it's the composition of `*` that is used to know the schema, and that is controlled by whether `hasMessageColumn` is true or not. But the other columns are optional.

Note that in the case of Oracle, most of them need to be optional, since Oracle does not distinguish between null and empty.

jroper:
You're right - I just checked, Akka persistence actually has a serializer defined for `PersistentRepr`, and does use Akka serialization to serialize the payload using the defined serializer for that payload. That makes this whole PR far less important than I had thought.

I think it's still of value to have it done this way - for example, if you want to manually read/manipulate the event log in any way, right now you need to know the protobuf format of PersistentRepr to do it. Also, not sure if Akka persistence intends for people to use this serializer for storage, they might just intend it for use in remoting, in which case, the only backwards compatibility that they'll be maintaining would be short term to allow for rolling upgrades. So from that perspective it's probably still pretty important, but quite as bad as I had thought.

skisel:
Hi there,

I'm wondering is there any plan to finish this and merge finally? May I help somehow?

src/test/scala/akka/persistence/jdbc/migration/V4MigrationSpec.scala
WellingR: Wouldn't it be possible to insert the rows in legacy format using slick? That would make it database independent

src/test/resources/schema/postgres/postgres-schema.sql
```diff
@@ -18,7 +22,9 @@ CREATE TABLE IF NOT EXISTS public.snapshot (
   persistence_id VARCHAR(255) NOT NULL,
   sequence_number BIGINT NOT NULL,
   created BIGINT NOT NULL,
-  snapshot BYTEA NOT NULL,
+  snapshot_DATA BYTEA NOT NULL,
```
WellingR: Is there any reason why DATA has been uppercased?

src/main/scala/akka/persistence/jdbc/journal/dao/JournalTables.scala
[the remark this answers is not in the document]
WellingR: As far as I know, some people use these table definitions to generate the ddl. (as opposed to the pre-defined ddl's that we use in the tests).

It is a fair point that oracle does not distinguish between null and empty, however to keep the behaviour the same for all database we could explicitly make the columns `eventManifest` and `serManifest` optional, and insert None/null in case the manifests are empty.

This allows the non-legacy journal (and the row case class) to omit the old `message` and the `event`, `serId` and `writerUuid` can be made mandatory.

It might require a bit of fiddling to get this right (since this gives us two table definitions), but I think it is worth it

WellingR:
I see that the biggest open issue is that the migration task still needs some improvement.
I would also prefer separate table definitions for the new/legacy journals.

@jroper are you planning to continue working on this soon?
If not, then I propose that I merge this PR to the 4.x.x branch and create some new issues from the work that still needs to be done to make this completely ready.

skisel:
As there is no response from @jroper, I would merge it to the branch and then somebody else could address these issues (I could try to contribute).

WellingR:
Merging this to the akka-serialization branch in this repository for now. This allows easier resolving of conflicts, and others can fix the remaining issues

WellingR:
@skisel If you want to address some of the open issues you can have a look at the new version of this PR #199 . Let me know if you want to have a go at one of the open issues

PR #181, opened 2018-04-10
use snapshot table name in its primary key name

notNotDaniel:
The default snapshot table schema used a fixed name (`snapshot_pk`) for the primary key, which makes it impossible to have more than one snapshot table in the same schema (since they both try to create the primary key with the same name). This change makes the name based on the configured table name, so that more than one can coexist.

Thanks.


notNotDaniel:
Good point, I noticed the snapshot table when we were implementing a custom journal table, so didn't notice the problem there too. I'll add the change for the journal table too.
Thanks.

PR #183, opened 2018-04-25
initial impl of in-place updates, needed for data migrations

ktoso:
Proposal for https://github.com/dnvriend/akka-persistence-jdbc/issues/182

It is not entirely complete yet as it will need some types from Akka itself.
This would allow us to write "encrypt all events" and other migrations on existing data


ktoso:
Mutating tags is left out on purpose, at least currently, we could get to it later

ktoso:
Lol, okey I can revert to -> and => in the code I added :-)

src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
```diff
@@ -62,6 +72,13 @@ class JdbcAsyncWriteJournal(config: Config) extends AsyncWriteJournal {
       case Failure(cause) => throw cause
     }
   }
+  // only accessed if we need to perform Updates -- which is very rarely
+  lazy val journalDaoWithUpdates: JournalDaoWithUpdates =
```
patriknw: can be a `def`

src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
```diff
@@ -87,7 +104,11 @@ class JdbcAsyncWriteJournal(config: Config) extends AsyncWriteJournal {
         // If the previous write failed then we can ignore this
         f.recover { case _ => () }.flatMap(_ => fetchHighestSeqNr())
     }
+  }
 
+  private def asyncUpdateEvent(persistenceId: String, sequenceNr: Long, message: AnyRef): Future[Unit] = {
+    system.log.info("Updating {}@{}, to: [{}]", persistenceId, sequenceNr, message)
```
patriknw: don't `toString` on unknown messages

src/test/scala/akka/persistence/jdbc/query/CurrentEventsByPersistenceIdTest.scala
```diff
@@ -133,6 +137,43 @@ abstract class CurrentEventsByPersistenceIdTest(config: String) extends QueryTes
...
+        tp.request(Int.MaxValue)
+          .expectNext(EventEnvelope(Sequence(1), pid, 1, 1))
+          .expectNext(EventEnvelope(Sequence(2), pid, 2, 2))
+          .expectNext(EventEnvelope(Sequence(3), pid, 3, 3))
+          .expectComplete()
+      }
+
+      // perform in-place update
+      val journalP = TestProbe()
+      journal.tell(JdbcAsyncWriteJournal.InPlaceUpdateEvent(pid, 1, Integer.valueOf(111)), journalP.ref)
```
patriknw: is the `Integer` because of `AnyRef` ?
wouldn't `111` be boxed anyway?

ktoso: Yeah had some issue there around that, will check again

src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
ktoso: Thanks!

src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
ktoso: ok

ignasi35:
> Mutating tags is left out on purpose, at least currently, we could get to it later

Very unfortunate. Would ease the fix for https://github.com/dnvriend/akka-persistence-jdbc/issues/168

ktoso:
One step at a time @ignasi35... :-)
Timing wise this week we can only manage this one.

ktoso:
Addressed the codacy bot comments anyway ;)

src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
```diff
@@ -34,9 +35,18 @@ import slick.jdbc.JdbcBackend._
...
 object JdbcAsyncWriteJournal {
   private case class WriteFinished(pid: String, f: Future[_])
+  /**
+   * Extra Plugin API: May be used to issue in-place updates for events.
+   * To be used only for data migrations such as "encrypt all events" and similar operations.
+   *
+   * The write payload may be wrapped in a [[akka.persistence.journal.Tagged]],
+   * in which case the new tags will overwrite the existing tags of the event.
+   */
+  final case class InPlaceUpdateEvent(pid: String, seqNr: Long, write: AnyRef)
```
WellingR: The parameter names can be improved. I would prefer renaming `pid` to `persistenceId`, and `seqNr` to `sequenceNr` (to keep the names consistent with the rest of the apis). Also I would suggest renaming `write` to something like `updatedPayload` or `replacementPayload` (or replacementMessage/replacementEvent, not sure what is best).

src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalDao.scala
```diff
@@ -107,6 +109,15 @@ trait BaseByteArrayJournalDao extends JournalDao {
       db.run(actions.transactionally)
     }
 
+  def update(persistenceId: String, sequenceNr: Long, payload: AnyRef): Future[Unit] = {
+    val write = PersistentRepr(payload, sequenceNr, persistenceId)
+    val serializedRow = serializer.serialize(write) match {
+      case Success(t)  => t
+      case Failure(ex) => throw new IllegalArgumentException(s"Failed to serialize ${write.getClass} for update of [$persistenceId] @ [$sequenceNr]")
+    }
+    db.run(queries.update(persistenceId, sequenceNr, serializedRow.message)).map(_ => ())
```
WellingR: Perhaps a `Future[Done]` can be returned, in that case we do not need to use `.map(_ => Done)` at a later point to convert it again.

WellingR:
Looks good so far.
Some comments/questions in general:
- It is probably good to add some documentation on how this can be used. However if this becomes a feature of akka-persistence then the akka-docs are probably the best place.
- The migration needed for #180 might actually be a good use case for showing that this works.

src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
ktoso: Good suggestions, thanks!

WellingR: Also, I was thinking, should the type of `write` be `Any` or `AnyRef`? In most other places the type of the message seems to be `Any`. Perhaps this also solved the boxing issues you mentioned in Patriks comment

src/main/scala/akka/persistence/jdbc/journal/dao/JournalDaoWithUpdates.scala
```diff
@@ -0,0 +1,30 @@
...
+
+package akka.persistence.jdbc.journal.dao
+
+import scala.concurrent.Future
+
+/**
+ * A [[JournalDao]] with extended capabilities, such as updating payloads and tags of existing events.
+ * These operations should be used sparingly, for example for migrating data from un-encrypted to encrypted formats
+ */
+trait JournalDaoWithUpdates extends JournalDao {
```
renatocaval: Do we really need it as a trait? Is the motivation behind it to not disturb users that are using their own implementation of `JournalDao`?

I personally don't think this should be a problem. If we add the method to `JournalDao`, users implementing their own DAO will get compiler error when upgrading. 

Unless they get it upgraded transitively, which is rather rare case, they will get a runtime error when using the method, but that is what they will get anyway with the current implementation.

src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
renatocaval: Yes, the type of the payload in`PersistentRepr` is `Any`. Although I agree that an event is supposed to be an `AnyRef`, the current API allows it to be `Any` (and as consequence also `AnyVal`).

renatocaval:
@ktoso there are still some change request and comments to address here. Could you have a look?

I think we can merge after that. LGTM.

src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalDao.scala
ktoso: done

src/main/scala/akka/persistence/jdbc/journal/dao/JournalDaoWithUpdates.scala
ktoso: I do think the separation is clear, not sure conflating journal api with this mutable terrible api it is helping

ktoso:
Applied the change. We do not want to document this too much, it for now is rather intended as a hidden "please don't use this" API, and we'll build tools around it that do end up using this internal api without showing people the full "power" of the update api.

Sounds good?

src/main/scala/akka/persistence/jdbc/journal/dao/JournalDaoWithUpdates.scala
renatocaval: Fine by me. 

I just don't find it useful to separate it. As as consequence we need to branch later on (like here https://github.com/dnvriend/akka-persistence-jdbc/pull/183/files#diff-381cd8e4dd090538bdd312eed8222f76R75).

Anyway, you don't need to change it. Just want to understand the motivation in creating a separated trait.

renatocaval:
@ktoso sounds good to me. No need to 'overdocument' it. :-)

 what about those comments: https://github.com/dnvriend/akka-persistence-jdbc/pull/183#pullrequestreview-116030853?

src/test/scala/akka/persistence/jdbc/journal/JdbcJournalSpec.scala
```diff
@@ -19,11 +19,11 @@ package akka.persistence.jdbc.journal
...
 import akka.persistence.jdbc.config._
 import akka.persistence.jdbc.util.Schema._
-import akka.persistence.jdbc.util.{ClasspathResources, DropCreate, SlickDatabase, SlickExtension}
+import akka.persistence.jdbc.util.{ ClasspathResources, DropCreate, SlickDatabase, SlickExtension }
 import akka.persistence.journal.JournalSpec
-import com.typesafe.config.{Config, ConfigFactory, ConfigValueFactory}
+import com.typesafe.config.{ Config, ConfigFactory, ConfigValueFactory }
 import org.scalatest.concurrent.ScalaFutures
-import org.scalatest.{BeforeAndAfterAll, BeforeAndAfterEach}
+import org.scalatest.{ BeforeAndAfterAll, BeforeAndAfterEach }
```
ktoso: does scalariform not run on compile? 
Or I guess after rebasing the settings changed?

renatocaval:
I have to check it, but got it after compiling locally

src/test/scala/akka/persistence/jdbc/journal/JdbcJournalSpec.scala
renatocaval: have you run `test:compile`? Most of the files that were reformatted were test files. Except one that was already in master and got merged here when I resolved the conflict

WellingR:
FYI, I made an attempt to disable to the codacy setting which disallows non ascii characters. So we should no longer get these warnings in the future.

PR #187, opened 2018-05-24
minor improvements on SlickDriver

renatocaval:
This PR adds back overloaded methods for `SlickDriver.forDriverName` and `SlickDriver.forConfig`. Those method got an extra param (path: String) in PR #151. 

That extra param is useless for those using JNDI lookup (eg: Lagom), so it's nice to have a variant without it and also to keep backward compatibility with previous versions.

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/util/SlickDriver.scala
+++ b/src/main/scala/akka/persistence/jdbc/util/SlickDriver.scala
@@ -25,14 +25,24 @@ import slick.jdbc.JdbcProfile
 import slick.jdbc.JdbcBackend._
 
 object SlickDriver {
+
+  def forDriverName(config: Config): JdbcProfile =
+    forDriverName(config, "slick")
+
   def forDriverName(config: Config, path: String): JdbcProfile =
     DatabaseConfig.forConfig[JdbcProfile](path, config).profile
 }
 
 object SlickDatabase {
-  def forConfig(config: Config, slickConfiguration: SlickConfiguration, path: String): Database = {
-    val jndiName = slickConfiguration.jndiName.map(Database.forName(_, None))
-    val jndiDbName = slickConfiguration.jndiDbName.map(new InitialContext().lookup(_).asInstanceOf[Database])
-    jndiName.orElse(jndiDbName).getOrElse(Database.forConfig(path, config))
-  }
+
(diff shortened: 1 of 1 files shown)


src/main/scala/akka/persistence/jdbc/util/SlickDriver.scala
```diff
@@ -25,14 +25,24 @@ import slick.jdbc.JdbcProfile
...
+  def forConfig(config: Config, slickConfiguration: SlickConfiguration): Database =
+    forConfig(config, slickConfiguration, "slick.db")
+
+  def forConfig(config: Config, slickConfiguration: SlickConfiguration, path: String): Database =
+    slickConfiguration.jndiName
+      .map(Database.forName(_, None))
+      .orElse{
+        slickConfiguration.jndiDbName.map(new InitialContext().lookup(_).asInstanceOf[Database])
+      }
+      .getOrElse(Database.forConfig(path, config))
```
renatocaval: Minor improvement. Only lookup JNDI resource if really needed.
 
I had added this to #150 that didn't make it because superseded by #151.

src/main/scala/akka/persistence/jdbc/util/SlickDriver.scala
```diff
@@ -25,14 +25,24 @@ import slick.jdbc.JdbcProfile
 import slick.jdbc.JdbcBackend._
 
 object SlickDriver {
+
+  def forDriverName(config: Config): JdbcProfile =
```
WellingR: While I understand that it is good to have a backwards compatible version of this method. I think this api is actually an internal api (even though it is not marked as such).

Would it be an idea to mark this method as deprecated?

The same holds for the other overload you added below.

renatocaval: I agree that this should be private. 

The problem now is that Lagom is using this API and because we use JNDI we don't have a path to pass.

Maybe we should add something similar in Lagom instead and not use the plugin one.

renatocaval: Let's try the following:

* Lagom get's its own JNDI lookup utility class
* the old methods are reintroduced in this PR, but deprecated
* we make the new ones private/internal API

How does it sound?

WellingR: Sounds good

PR #193, opened 2018-06-05
Use system.registerOnTermination instead of Coordinated shutdown

### LINKED ISSUE #192: When using shared db's connections may be left open.
NOTE: `akka-persistence-jdbc` distinguishes between `dbs` and `shared-dbs`

* `akka-persistence-jdbc` uses `Actor#postStop` in `JdbcAsyncWriteJournal` and  `JdbsSnapshotStore`. There, `db.close` is invoked only if `db` is not shared. 
* `akka-persistence-jdbc` uses `CoordinatedShutdown#addTask` in `JdbcReadJournal`. `JdbcReadJournal` uses `coordinated-shutdown-phase`.
* `SlickExtension`'s `DbHolder` uses `CoordinatedShutdown#addTask` registering a task every time a new `db` is provided. `SlickExtension` uses `shared-db-coordinated-shutdown-phase`
* Both `coordinated-shutdown-phase` and `shared-db-coordinated-shutdown-phase` default to the phase `before-actor-system-terminate`.

`CoordinatedShutdown` only runs if invoked. By default `CoordinatedShutdown` terminates the `ActorSystem` and it is run on `SIGTERM`.

#### The problems:

1. Using the phase `before-actor-system-terminate` instead of the phase `actor-system-terminate` means actors are stopped _after_ `db.close` is run.
1. When an `ActorSystem` is _terminated_  using `ActorSystem.terminate` (as opposed to _shut down_ using `CoordinatedShutdown.run`), then tasks registered in `CoordinatedShutdown` are not run and `db`'s may be left open. There are scenarios where users simply use `ActorSystem.terminate` (e.g. tests). It is also possible users have their own JVM shutdown hook invoking `ActorSystem.terminate` in which case there's a race condition to see which JVM shutdown hook (userland v Akka's JVM shutdownhook) will invoke `db.close` first).

#### Proposals

1. Fixing this issue could be as simple as adding more docs: "in some scenarios only  `CoordinatedShutdown.run` will release all resources whihle `ActorSystem.terminate` won't" (needs nicer wording).
2. As a library, `akka-persistence-jdbc` may not rely on `CoordinatedShutdown` and instead trust only the actor's `postStop` and the ActorSystem's `registerOnTermination` features. Coordinated Shutdown will run `ActorSystem.terminate` in the last phases anyway.

WellingR:
Fixes #192 

I have removed all usages of and references to coordinates shutdown and replaced them with system.registerOnTermination.

The existing configuration to add a shutdown hook is still used. So it is still possible to disable this functionality


WellingR:
@renatocaval 

> SlickExtensions, always creates the db being it shared or not

Not completely true: the `DbHolder` for a shared database is created when the default `SlickExtension` is loaded. The `DbHolder` uses a `lazy val` to ensure that the database is created at most once.
Non-shared database are always created when requested.

Concerning your suggestion

> DbHolder doesn't need to have an addShutDownHook boolean. The rule being, if there is a DbHolder, it's shared and therefore we need to close it on system shutdown, not the component. 

Does this mean you want to drop the `shared-db-add-shutdown-hook` setting? I actually agree that it probably makes sense to always shut down a shared database, if a user wants to shutdown the database manually it is still possible do configure a custom implementation of the slickextension.


> When storing the DB reference in JNDI (Lagom's case), the DB is shared, but the plugin makes no distinction from a non-shared one.
We don't need to add it on this PR, but we could also use a DbHolder for the JNDI Db. In which case, when shutting down the actor system, it will only be closed once and not per component.

Actually all we need to do is to document that JNDI databases should be configured as shared databases. Perhaps we could even log a warning if JNDI is used in a non-shared database.

renatocaval:
@WellingR 

> SlickExtensions, always creates the db being it shared or not

What meant by that is that creation is managed by the extension. It's the extension that always make the decision. For instance, it makes the decision that a Db will be created lazily by the `DbHolder`. It can be the one holding all the necessary information and signaling to the other parts how they should behave.

> Does this mean you want to drop the shared-db-add-shutdown-hook setting?

Yes, I think it's not needed anymore and users in need of a more fine grained control can impl their own variant. 

--- 

With respect to the JNDI case, I don't know what is the best approach. 
In the specific case of Lagom, the JNDI Db is configured somewhere else. When the plugin bootstraps and finds nothing on the normal config path, it tries a lookup in JNDI. At that moment, the JNDI Db was already created and it's waiting for the plugin to come pick it.

In that case, it's not possible to configure it as a shared Db, because it will be wrapped by a `DbHolder` but nobody will add it to the JNDI.

My initial thought was to detect that we were using a JNDI and treat is as a shared Db. We can make the extension aware that it's using something from JNDI. In that case, it returns the JNDI Db, but it register a shutdown hook for it (like with shared db). And it signals to the components that they should not close the Db if they crash for some reason.

Or even better, if we detect that we are using JNDI, we tell the component to NOT close it, but the we DON'T register a shutdown hook neither. It's the one that created the Db and register it in JNDI the responsibility to close it, not the plugin. The plugin is just consuming a JNDI resource, it's not responsible to managing it. 

We can leave the JNDI case on the side for now. Let's think more about it and let's not block this current PR because of it.

WellingR:
@renatocaval I have implemented the changes you suggested. Can you have a look?

src/main/scala/akka/persistence/jdbc/util/SlickDatabase.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+   * If true, the requesting side usualy a (read/write/snapshot journal)
+   * should shutdown the database when it closes. If false, it should leave
+   * the database connection pool open, since it might still be used elsewhere.
+   */
+  def allowShutdown: Boolean
+}
+
+case class StrictSlickDatabase(database: Database, profile: JdbcProfile) extends SlickDatabase {
```
WellingR: I am not really sure if this is the best name for the "simple" version if the SlickDatabase

WellingR:
By the way, a disadvantage of the updated implementation is that it breaks the SlickExtension interface which we released in the previous version.

src/main/scala/akka/persistence/jdbc/util/SlickDatabase.scala
renatocaval: Maybe we should call it `EagerSlickDatabase` instead as opposed to the `lazy` one.

src/main/scala/akka/persistence/jdbc/util/SlickDatabase.scala
```diff
@@ -0,0 +1,118 @@
...
+      profile(config, path))
+  }
+}
+
+trait SlickDatabase {
+  def database: Database
+  def profile: JdbcProfile
+
+  /**
+   * If true, the requesting side usualy a (read/write/snapshot journal)
```
renatocaval: typo: usually

renatocaval:
Yes, can you do it? I'm out until next week.

Issue #172, opened 2018-03-14
Include a configuration setting which ensures that events can only have a single tag, for better eventsByTag performance


WellingR:
See the discussion in #168

> I'd suggest that we have the config disable multiple tag support entirely, fail fast when persisting multiple tags, and maybe warn or even fail when deserializing events with multiple tags.

Since we can do a simple equality check if we only use a single tag, this might make the query perform better.

agnaldo4j:
Hi @WellingR  at this moment my team has a performance problem with that eventsByTag implementation, we try to use some Postgres extensions (https://www.postgresql.org/docs/9.6/static/pgtrgm.html) to help but for us a query like this ``` tags = 'teg' ``` would be great.

WellingR:
Hey @agnaldo4j,
To fix the problem that you have right now, you can create your own implementation of 
`ReadJournalDao` (which you can derive from `ByteArrayReadJournalDao`) and configure it your application.conf.

agnaldo4j:
Hi @WellingR,

I Implement like you said and it improves the performance of the database.

Why you prefer this approach that uses LIKE in default mode? what is the most common use case, to use it in the views?

thanks for your help.

WellingR:
@agnaldo4j 
It is not my preference to use the LIKE query, however this is still how it is implemented right now. The main reason is that akka-persistence allows storing multiple tags for an event.

The proper solution could be to store tags in a separate table (then it is still possible to store multiple tags). It is on my todo list to investigate what the performance implications are of doing this.
In case I find big disadvantages of storing the tags separately, then I guess using a configuration setting for storing either a single tag, or multiple tags could be an option.

Issue #168, opened 2018-03-01
eventsByTag query returns incorrect results for tags that share a prefix


TimMoore:
The query for `eventsByTag` uses a `LIKE` to do a substring match on the provided tag:

https://github.com/dnvriend/akka-persistence-jdbc/blob/8a3ad9556ca5ccfbaec11f2b5c3bc39b39bd054a/src/main/scala/akka/persistence/jdbc/query/dao/ByteArrayReadJournalDao.scala#L97-L110

This means that if you use tags that contain a shard number, you'll have conflicts with more than single-digit shards. For example, when there is a tag such as `JobEvent12`, it will be matched by queries for `JobEvent1`.

renatocaval:
I can see only two solutions here:

1. add a flag to swap between `=` and `LIKE` syntax. 
2. add more tag columns and use `=` syntax, ie: "tag1 =  'JobEvent1' OR tag2 = 'FooBar'"

Option 1 is a damm hack and will also force people to stick with one tag only. 

Option 2 is better but is backward incompatible. 

Any other more creative ideas?

TimMoore:
I was thinking something like this:

```scala
val startsWithTag = s"$tag,%"
val endsWithTag = s"%,$tag"
val containsTag = s"%,$tag,%"

// ...
    WHERE (
      "#$tags" = $tag OR
      "#$tags" LIKE $startsWithTag OR
       "#$tags" LIKE $endsWithTag OR
      "#$tags LIKE containsTag
    ) AND
// ...
```

I don't know if that's going to cause terrible query performance, however. It might be better to have an exploded tags table with a foreign key relation to the persistence ID that is written in a transaction with the journal table, then join them to query. That of course would be a massive change requiring a schema migration.

TimMoore:
Possible that the query already requires a full table scan, however... I've never checked.

renatocaval:
> val startsWithTag = s"$tag,%"
val endsWithTag = s"%,$tag"
val containsTag = s"%,$tag,%"

That's indeed creative! I like it. :-) 

Not sure about the performance implications of it.

TimMoore:
You can tell I spent some past time as a Perl programmer :joy:

bpiper:
[...]

PR #196, opened 2018-07-11
WIP: store tags in a separate table

WellingR:
As an experiment. I wanted to make the changes needed to store tags in a separate table. Here is my work so far.

- Since this first version is just to find out if this works, I decided to not yet make this backwards compatible. In case we do choose to continue with this, we should definitely ensure backwards (and forwards) compatibility. Also we need a simple way to migrate.
- There is one H2 test that seems to fail with timeouts, I have not yet figured out why this fails for H2 bt not for other databases
- I did not yet execute performance tests. 
  - We now insert in two tables, so inserting will be slightly slower (we should test how much)
  - The eventsByTag query needs to search in two tables, and the filtering conditions (tag and ordering) are in two different tables. So I guess the query strategy will still not be optimal (but hopefully better than before).

If we complete this work then this PR will solve #168 and it will make #172 obsolete

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -124,6 +124,15 @@ jdbc-journal {
         message = "message"
       }
     }
+    journal-tag {
+      tableName = "journal_tag"
+      schemaName = ""
+      columnNames {
+        tag = "tag"
+        persistenceId = "persistence_id"
+        sequenceNumber = "sequence_number"
+      }
+    }
   }
 
   tagSeparator = ","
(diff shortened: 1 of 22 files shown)


WellingR:
I am also investigating other options, such as the suggestion I made as part of #172 and the suggestion to use an array or jsonb to store multiple tags (which only works in some databases).

renatocaval:
@WellingR, I would like to re-consider the backwars / forwards compatibility. See #318. 

In a nutshell, we don't think that the efforts to make a seamless migration is worth the extra coding effort and code complexity.

renatocaval:
@WellingR, I'm closing this one (as discussed in #318)

PR #197, opened 2018-09-01
Documentation fixes

WellingR:
- Fixed a mistake in the configuration example for custom database providers
- Improved documentation on thread/connection pool sizing (specifically by removing the vague 5 * number of cores recommendation and pointing to a better resource)

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -58,7 +58,6 @@ akka-persistence-jdbc {
         # driver = "org.postgresql.Driver"
 
         # hikariCP settings; see: https://github.com/brettwooldridge/HikariCP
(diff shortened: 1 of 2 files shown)


src/main/resources/reference.conf
```diff
@@ -98,13 +97,10 @@ akka-persistence-jdbc {
         # ensures that the database does not get dropped while we are using it
         keepAliveConnection = on
 
-        # 5 * number of cores
+        # See some tips on thread/connection pool sizing on https://github.com/brettwooldridge/HikariCP/wiki/About-Pool-Sizing
+        # Keep in mind that the number of threads must equal the maximum number of connections.
```
renatocaval: we should eventually deprecate it and only have `maxConnections`. Not something to do right now, though

PR #199, opened 2018-09-02
WIP: Added support for serializing using Akka serialization

WellingR:
This is a continuation of #180

Todos (help wanted):

- [x] Improve migration script [see here](https://github.com/dnvriend/akka-persistence-jdbc/pull/180#pullrequestreview-112228857) and [here](https://github.com/dnvriend/akka-persistence-jdbc/pull/180#discussion_r181574113) (in progress: #201)
- [x] Investigate if we can split out the legacy and new table definitions [see here](https://github.com/dnvriend/akka-persistence-jdbc/pull/180#discussion_r209034994) (in progress #203)
- [ ] Make sure that the tests use both the new and the legacy configurations.
- [x] Re-enable the `supportsSerialization` compatibility flag (the tests break if this is enabled)
- [ ] The journal migration tests should work for all supported types of databases. (Perhaps we could read the data from a csv file and convert that to some slick insert statements or something).
- [ ]  The journal migration uses a `RowsPerTransaction` configuration value which is currently hardcoded. It might be an idea to include `rowsPerTransaction` as a method parameter when the migration is started.
- [ ]  The changes from #200 have been merged to master, these changes have been merged to this branch too, however there is no schema migration script yet. We should also investigate if additional tests are needed.
- [ ]  Fix oracle sql schema migration bug [see here](https://github.com/dnvriend/akka-persistence-jdbc/pull/199#issuecomment-422555621)


renatocaval:
Thanks @WellingR. Good idea to open a serialization branch for this.

src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalSerializer.scala
```diff
@@ -17,28 +17,72 @@
...
+      event <- tryEvent
+    } yield {
+      val serializer = serialization.findSerializerFor(payload)
+      val serManifest = serializer match {
+        case stringManifest: SerializerWithStringManifest =>
+          stringManifest.manifest(payload)
+        case _ if serializer.includeManifest =>
+          payload.getClass.getName
+        case _ => ""
+      }
```
patriknw: this should be replaced by
```
val serManifest = akka.serialization.Serializers.manifestFor(serializer, payload)
```

skisel:
It looks to me slick has a bug here: https://github.com/slick/slick/blob/master/slick/src/main/scala/slick/jdbc/OracleProfile.scala#L155

Slick generates something like this for migration code (I shortened) and on non-migrated db it should return ALL rows, but no result returned:
```select * from "journal" where dbms_lob.compare("event", NULL) = 0)``
This actually works:
```select * from "journal" where "event" is null```

By changing from ```.filter(_.event.isEmpty)``` to ```.filter(_.serId.isEmpty)``` it actually works. I will try to improve migration and will submit PR.

skisel:
https://github.com/dnvriend/akka-persistence-jdbc/pull/201

skisel:
Another bug to add here: 
`sql> ALTER TABLE "journal" MODIFY "message" BLOB DEFAULT NULL
[2018-09-18 23:08:08] [99999][22859] ORA-22859: invalid modification of columns`

PerWiklander:
It looks like the things left to do here are related to migrations. What is the status of this feature for new projects without old events to support?

WellingR:
@renatocaval @skisel 
This PR has been silent for almost a year, it is now behind master and there are probably some conflicts that may need to be resolved.

Do we still want to make these improvements even given the fact that migrations are needed?
Is any contributor willing to spend the time to work on finishing the remaining work?

I would be happy to contribute myself, but time is an issue.

renatocaval:
@WellingR, we have talked about this with the Akka team recently. We will need to make time to move one with that. We don't have it yet planned, but it's in our radar. 

Not only that but that tagging issue. Actually, tagging is the most important one.  And because we need some migration tooling, it will be a good opportunity to tackle that one altogether.

renatocaval:
Also closing this one as we will be taking a different approach as discussed in #318

PR #108, opened 2017-06-14
Provide support for MS-SQL-server

naferx:
Ref #102

PR #200, opened 2018-09-05
Add support for SQL Server

jonfox:
Adds SQL Server schema and test updates. I created a derived SQL server [image](https://hub.docker.com/r/topaztechnology/mssql-server-linux/) that allows login and database creation on startup to facilitate testing.

Write performance was initially poor vs Postgres, but significantly improved with use of `transactionally` in  `BaseByteArrayJournalDao.writeJournalRows`

This was based on a closed PR #108, thanks to @naferx

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -17,7 +17,9 @@ done;
 }
 rm ./bintray.sbt
 
+wait 5432 PostgreSQL
 wait 3306 MySQL
 wait 1521 Oracle
+wait 1433 SqlSever
 
 sbt test
--- a/src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalDao.scala
+++ b/src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalDao.scala
@@ -17,8 +17,6 @@
 package akka.persistence.jdbc
 package journal.dao
 
-import java.io.NotSerializableException
-
 import akka.{ Done, NotUsed }
 import akka.persistence.jdbc.config.JournalConfig
 import akka.persistence.jdbc.serialization.FlowPersistentReprSerializer
@@ -28,7 +26,7 @@ import akka.stream.scaladsl.{ Keep, Sink, Source }
 import akka.stream.{ Materializer, OverflowStrategy, QueueOfferResult }
 import org.slf4j.LoggerFactory
 import slick.jdbc.JdbcBackend._
-import slick.jdbc.JdbcProfile
+import slick.jdbc.{ JdbcProfile, H2Profile }
 
 import scala.collection.immutable._
 import scala.concurrent.{ ExecutionContext, Future, Promise }
@@ -86,9 +84,14 @@ trait BaseByteArrayJournalDao extends JournalDaoWithUpdates {
     }
   }
 
-  private def writeJournalRows(xs: Seq[JournalRow]): Future[Unit] = for {
-    _ <- db.run(queries.writeJournalRows(xs))
-  } yield ()
+  private def writeJournalRows(xs: Seq[JournalRow]): Future[Unit] = {
+    if (slickProfile != H2Profile)
+      // Write atomically without auto-commit
+      db.run(queries.writeJournalRows(xs).transactionally).map(_ => Unit)
+    else
+      // However transactionally causes H2 tests to fail
+      db.run(queries.writeJournalRows(xs)).map(_ => Unit)
+  }
 
   /**
    * @see [[akka.persistence.journal.AsyncWriteJournal.asyncWriteMessages(messages)]]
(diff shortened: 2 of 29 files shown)


WellingR:
Awesome! Perhaps it is an idea to merge this after #199.

In that case we do not need support for the legacy schema for sql server.

jonfox:
Yes I've noticed there seem to be intermittent failures on the 2.11 build. I have a Travis build hooked up to my fork and it passed maybe 50% of the time, so presumably load related. I haven't had a chance to investigate yet.

Also Codacy seems to be incorrectly complaining about valid T-SQL schema, not sure if it is possible to tag the SQL dialect.

In terms of merging once everything is reviewed and behaving, I have a fairly prompt requirement so would prefer sooner rather than later if at all possible, not sure what the timeline is for #199. I'd be happy to work on any updates required as part of that PR.

WellingR:
It might take a while before #199 is completely ready. So perhaps we can merge this one before.

The unstable travis build is a blocker though. This needs some investigation

jonfox:
Sure I will take a look at this and get it passing reliably on Travis

jonfox:
It looks like rebasing on top of your `dependency-updates` branch, and updating mssql-jdbc to 7.0.0 has Travis passing now.

jonfox:
or maybe not ... back to the drawing board.

jonfox:
I think the tests are passing reliably now, I've run Travis hooked up to my repo 10 times against the latest commit successfully.

The problem seemed to be down to the `transactionally` change; it looks like H2 doesn't like this when writing large batches of messages, so I have excluded use with the H2 profile. Of course it's preferable the write should be atomic, and SQL Server performance is very poor otherwise.

src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalDao.scala
```diff
@@ -86,9 +84,14 @@ trait BaseByteArrayJournalDao extends JournalDaoWithUpdates {
...
-  private def writeJournalRows(xs: Seq[JournalRow]): Future[Unit] = for {
-    _ <- db.run(queries.writeJournalRows(xs))
-  } yield ()
+  private def writeJournalRows(xs: Seq[JournalRow]): Future[Unit] = {
+    if (slickProfile != H2Profile)
+      // Write atomically without auto-commit
+      db.run(queries.writeJournalRows(xs).transactionally).map(_ => Unit)
+    else
+      // However transactionally causes H2 tests to fail
+      db.run(queries.writeJournalRows(xs)).map(_ => Unit)
```
renatocaval: Why H2 is failing when using `transactionally`? That's not normal.

src/test/scala/akka/persistence/jdbc/journal/JdbcJournalSpec.scala
```diff
@@ -62,6 +62,7 @@ class PostgresJournalSpecSharedDb extends JdbcJournalSpec(ConfigFactory.load("po
 class PostgresJournalSpecPhysicalDelete extends JdbcJournalSpec(ConfigFactory.load("postgres-application.conf")
   .withValue("jdbc-journal.logicalDelete", ConfigValueFactory.fromAnyRef(false)), Postgres())
 
+class MySQLJournalSpec extends JdbcJournalSpec(ConfigFactory.load("mysql-application.conf"), MySQL())
```
renatocaval: Wow! That one was missing and nobody noticed it. Thanks

build.sh
```diff
@@ -17,7 +17,9 @@ done;
 }
 rm ./bintray.sbt
 
+wait 5432 PostgreSQL
```
renatocaval: hmm, PostgreSQL was not included by default?

src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalDao.scala
renatocaval: I forgot I had added this comment and merge it. 

In any case, I don't think we are doing any harm here. Actually, I run this locally without the if/else and calling `transactionally`. All H2 tests passed. 

Moreover, we don't need it transactionally because we have only one `DBIO`. `transactionally` is only needed when we have many DBIOs and we want to run them all together in the same tx. That's not the case here. Here we have one `DBIO` with many row inserts.

jonfox: Yes the test failures were intermittent. Locally I also didn't have a problem, but maybe the Travis hardware was lower spec and caused the circuit breaker to time out. It's also not clear why this would reduce performance in H2 in the first place.

With regards to not requiring `transactionally`, it certainly made a significant difference to SQL Server performance. It's calling `JournalTableC ++= xs.sortBy(_.sequenceNumber)` but I have no idea how Slick handles this internally.

PR #203, opened 2018-09-09
Investigation if the legacy and new table definitions can be split out

skisel:
Here is my very draft of address #199 second point. It looks to me it does not worth an effort or requires to introduce some more complexity. 

It seems abstracting on the Table definition is not possible because of slick complexity with typing/type aliases/etc, so it will be only possible on DAO level, end solution would have 2 DAO's - one for legacy, another one for new journal. Same goes for Read-side DAO. It looks to me it will be a lot of copy-paste as well, as I was not able to abstract over queries. 

While doing this exercise I started to think that the best would be to release 2 versions - one for migration and backward/forward compatibility and another one for those who start from scratch or who already migrated their journals. Migration version could be supported for a while and then dropped at some point.

Let me know your opinion or help me with the direction. I am happy to continue if you agree it is worth it.

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalDao.scala
+++ b/src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalDao.scala
@@ -17,8 +17,6 @@
 package akka.persistence.jdbc
 package journal.dao
 
-import java.io.NotSerializableException
-
 import akka.{ Done, NotUsed }
 import akka.persistence.jdbc.config.JournalConfig
 import akka.persistence.jdbc.serialization.FlowPersistentReprSerializer
@@ -134,7 +132,7 @@ trait BaseByteArrayJournalDao extends JournalDaoWithUpdates {
       case Success(t)  => t
       case Failure(ex) => throw new IllegalArgumentException(s"Failed to serialize ${write.getClass} for update of [$persistenceId] @ [$sequenceNr]", ex)
     }
-    db.run(queries.update(persistenceId, sequenceNr, serializedRow.message, serializedRow.event, serializedRow.serId, serializedRow.serManifest).map(_ => Done))
+    db.run(queries.update(persistenceId, sequenceNr, serializedRow).map(_ => Done))
   }
 
   private def highestMarkedSequenceNr(persistenceId: String) =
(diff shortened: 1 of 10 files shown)


WellingR:
Releasing separate versions could be an option, however in this case v4.0.0 will be the "legacy" version. and in v5.0.0 we could drop the support for the old journal.

I did some experimentation and I believe we still have some possibilities. Please have a look at
https://gist.github.com/WellingR/2c9892e8cba87364b6292c792face0c2
This method might work for the queries, but not for the insert/delete/update actions.

Perhaps, with some clever use of generics, the Daos/queries can get the type of the TableQuery and je JournalRow as a type parameter. that way we might be able to re-use the same dao code.

skisel:
This is what I failed to achieve actually, thanks for the hint. Will try to explore it a bit further.

skisel:
Please have a look to second try. Not much reusable code by using provided approach, but there is some.

WellingR:
I agree that the amount of code reuse is very limited. The orginal proposal by jroper had much less impact on the daos.

One of the reasons for the separation would be to allow users to generate the ddl based on the table class. I guess the best way to have both optimal code reuse AND the ability to generate the ddl might be the following.

Based on the value of the `hasMessageColumn` boolean we can determine if the table ddl is in the legacy mode, or in the new mode. Based on this, the current implementation already decides whether to include the message column or not.

In addition we can do same for the columns which are no longer optional in the new format. We could do this as follows:
```scala
    val event: Rep[Option[Array[Byte]]] =
      if (journalTableCfg.hasMessageColumn) column[Option[Array[Byte]]](journalTableCfg.columnNames.event)
      else Rep.Some(column[Array[Byte]](journalTableCfg.columnNames.event))
    val serId: Rep[Option[Int]] =
      if (journalTableCfg.hasMessageColumn) column[Option[Int]](journalTableCfg.columnNames.serId)
      else Rep.Some(column[Int](journalTableCfg.columnNames.serId))
    val writerUuid: Rep[Option[String]] =
      if (journalTableCfg.hasMessageColumn) column[Option[String]](journalTableCfg.columnNames.writerUuid, O.Length(36, varying = true))
      else Rep.Some(column[String](journalTableCfg.columnNames.writerUuid, O.Length(36, varying = true)))
```

A quick test `JournalTable.schema.createStatements.foreach(println)` makes it seem that depending on the `hasMessageColumn` value, the ddl generated by slick changes indeed.

@skisel What do you think?

WellingR:
Yes, the case class definition will indeed stay without change, but that should not be a big problem.
I would appreciate it if you could do this completely (also for the snapshot table)

skisel:
Ok, I will create another pull request with complete feature then. Closing this one

PR #204, opened 2018-09-13
Use systemActorOf to allow use with Akka Typed

natevecc:
Switch the JdbcReadJournal actor to be created as a system actor so it can be used with Akka Typed. The same thing had to be done in this [library](https://github.com/dnvriend/akka-persistence-inmemory/pull/54). 

I did a search of the entire code base to make sure `actorOf` isn't used anywhere else. It is used in the tests but I didn't change those since it doesn't impact users.


AndreasPresthammer:
Any plans on releasing this fix any time soon?

WellingR:
I will make sure we release a new version once #207 has been resolved.

AndreasPresthammer:
Thanks for quick reply :)

Any guess on an ETA on that?

PR #205, opened 2018-09-18
#199 WIP. Fixing DDL generation.

skisel:
#199 
Fixed DDL generation for legacy and non-legacy journal/snapshot table definitions.
Rows per transaction has been made configurable.
event_manifest column has been removed as it's duplicated with ser_manifest in the journal.

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -16,6 +16,11 @@ akka-persistence-jdbc {
 
   migration {
     v4 {
+      #Number of rows to be migrated in the journal per single transaction in the database
+      journal.rows-per-transaction = 100
+      #Number of rows to be migrated in the snapshot store per single transaction in the database
+      snapshot.rows-per-transaction = 100
+
       # Enables backwards compatibility when migrating to version 4.x, to ensure that any old events or snapshots
       # persisted by akka-persistence-jdbc 3.x can still be read by akka-persistence-jdbc 4.x. This must be enabled
       # if you have data in your database that was produced by akka-persistence-jdbc 3.x, otherwise that data will not
--- a/src/test/resources/schema/oracle/oracle-migration-v4.sql
+++ b/src/test/resources/schema/oracle/oracle-migration-v4.sql
@@ -1,4 +1,3 @@
-ALTER TABLE "journal" MODIFY "message" BLOB DEFAULT NULL;
 ALTER TABLE "journal" ADD (
   "event" BLOB DEFAULT NULL,
   "event_manifest" VARCHAR(255) DEFAULT NULL,
@@ -7,7 +6,6 @@ ALTER TABLE "journal" ADD (
   "writer_uuid" VARCHAR(36) DEFAULT NULL
 );
 
-ALTER TABLE "snapshot" MODIFY "snapshot" BLOB DEFAULT NULL;
 ALTER TABLE "snapshot" ADD (
   "snapshot_data" BLOB DEFAULT NULL,
   "ser_id" INTEGER DEFAULT NULL,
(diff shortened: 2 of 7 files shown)


src/main/resources/reference.conf
```diff
@@ -12,10 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+akka-persistence-jdbc.migration.v4.journal.rows-per-transaction = 100
+akka-persistence-jdbc.migration.v4.snapshot.rows-per-transaction = 100
```
WellingR: This is duplicated with the configuration below, so these can be removed

skisel: I thought it would be convenient to limit rows per transaction for journal and snapshots as the average size of event and snapshot might be different, so speed of migration can be different.

WellingR:
Have a look at `akka.persistence.journal.EventAdapter`.

By looking for usages of `PersistentRepr.manifest` I see that this manifest is indeed used for the event adapter. What is odd is that the tests still pass. Perhaps the akka-persistence-tck is missing a testcase?

src/main/resources/reference.conf
WellingR: I agree. However what I meant to say what these configuration settings are also set on line 22 and 24, so there is no need to configure this twice

skisel:
It was confusing for me that Snapshot does not have both, so I thought probably only one is required. I checked akka-persistence code and I see that during creation of PersistentRepr.manifest is not being assigned:

[Eventsourced.scala](https://github.com/akka/akka/blob/master/akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala#L349)

Then plugins concern is to take Event of type T and transform it to Array[Byte] using serializer and get manifest from the serializer. Then it is required for reading side by event adapter (when it is already deserialized to T)

src/main/resources/reference.conf
skisel: ah you're right, committed by accident :)

WellingR:
I think event adapters are only supported for events, but not for snapshots. Which is why the event adapter manifest is only relevant for events.

From the akka-persistence code it seems that the manifest is set at a later point. See
[WriteJournalBase.scala](https://github.com/akka/akka/blob/cc0705483b0860f258e8d11d6e315acd5749cf41/akka-persistence/src/main/scala/akka/persistence/journal/WriteJournalBase.scala#L43)

skisel:
I still dont get how on the persistence level there could be two manifests? what does it mean to have one set to "a" and another "b"?

WellingR:
There are serializers and event adapters.
See https://doc.akka.io/docs/akka/2.5/persistence.html#event-adapters for some explanation about event adapters

WellingR:
@patriknw can you confirm that the `manifest` from the `PersistentRepr` is different from the manifest of the serializer?

skisel:
Can we ping @patriknw somehow differently? :) I dont want to continue on this point unless he can confirm.

patriknw:
@WellingR @skisel The manifest in `PersistentRepr` has to do with event adapters, not serialization of the payload.

skisel:
@patriknw should it be persisted as well so PersistentRepr can be unserialised?

patriknw:
yes, it's needed in the event adapter `fromJournal` when reading

skisel:
thanks @patriknw 

@WellingR I will revert removal of this manifest.

skisel:
@WellingR please review

skisel:
anyone to review? please review, i want to finalize it

src/test/resources/schema/oracle/oracle-migration-v4.sql
```diff
@@ -1,4 +1,3 @@
-ALTER TABLE "journal" MODIFY "message" BLOB DEFAULT NULL;
```
WellingR: Should this be deleted? Why?
(same for the snapshot column in the snapshot table)

skisel: it does not work:
```
sql> ALTER TABLE "journal" MODIFY "message" BLOB DEFAULT NULL [2018-09-18 23:08:08] [99999][22859] ORA-22859: invalid modification of columns
```

skisel:
so oracle migration does not work properly and needs to be adjusted. I suggest to tackle it as separate PR and merge this one into the branch.

PR #210, opened 2018-10-05
Adds allowPublicKeyRetrieval=true to all MySQL URLs.

odd:


renatocaval:
Hi @odd. can you explain the motivation for this change?

odd:
Hi @renatocaval, I should have provided a description when creating the PR, but I forgot. 

When I cloned the repository and ran the test suite locally, the MySQL tests did not execute but instead they all hanged for a while and then failed with `java.sql.SQLNonTransientConnectionException: Public Key Retrieval is not allowed`. After adding `allowPublicKeyRetrieval=true` to the MySQL JDBC URL:s the problem went away. 

For reference, I'm using a MacBook Pro (Retina, 13-inch, Early 2015) using MacOS Mojave version 10.14.1 Beta (18B57c).

odd:
No, I run it with Docker (using the `test.sh` script). Perhaps I have some leftovers of old MySQL installations hanging around on my computer; or perhaps it has to do with the Docker for Mac installation (Docker Community Edition Version 2.0.0.0-beta1-mac75 (27117), Engine: 18.09.0-ce-beta1, )... 🤔

renatocaval:
Ok, good for me. Let's merge it because there is no harm.

renatocaval:
Thanks @odd

PR #215, opened 2019-01-10
Ensure only one row is returned in appropriate snapshot queries

richbirch:
We recently found that recovering an actor with large sized snapshots failed due to a missing limit on the query. The query as is stands returns every snapshot meeting the criteria which for us was several 10s of GBs of data.

There was an existing limit on  `_selectByPersistenceIdAndMaxSeqNr`, but none on the other 3 similar queries where a max sequence number or timestamp are specified:

- _selectByPersistenceIdAndMaxTimestamp
- _selectByPersistenceIdAndMaxSequenceNr
- _selectByPersistenceIdAndMaxSequenceNrAndMaxTimestamp


WellingR:
Thanks for contributing!
As you might have seen from the Travis build, some of the tests are failing.
It seems that the queries you have changed are not only used for retrieving the data, but they are also used as the basis for the delete queries (for cleaning up the snapshot data).

My suggestion would be to split up the queries, so that we have separate queries for retrieving the data, and separate queries which can be used for deleting the data.

richbirch:
Apologies for not having checked the tests. I actually thought we had checked them locally but clearly we hadn't. I'll update the PR once I've had a little time to work on it

WellingR:
@richbirch Do you think you have the time to resolve this soon? If not, then I am willing to fix this myself.

PR #217, opened 2019-01-20
travis settings and h2 transactionally

WellingR:
This PR includes the first commit of #207 so that the writing of journal rows always runs in a transaction regardless of which database is used.

~~It turn out that H2 needs a bit more memory, however from some of my tests, it seemed that the memory settings we try to pass in the .sbtopts file are overridden by travis. To work around this I pass the memory parameters to SBT explicitly~~

It seems that the issues we are having with h2 only occur in h2 version 1.4.197 and not in the previous version. So I downloaded h2 to 1.4.196. (many issues seem to have been reported and fixed on 197, but no release has been made yet for 9 months).

In addition I discovered another bug in our build, namely that the scala 2.11 build was actually a scala 2.12 build... This has been fixed too.

Edit: 
Somehow our oracle image got deleted from the docker hub again, however the one that we used half a year ago is back again. So I switched back to that oracle image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ services:
   - postgresql
   - docker
 before_install:
-  - docker pull alexeiled/docker-oracle-xe-11g
+  - docker pull wnameless/oracle-xe-11g
   - docker pull topaztechnology/mssql-server-linux
-  - docker run --name oracle -d -p 127.0.0.1:1521:1521 -e ORACLE_ALLOW_REMOTE=true alexeiled/docker-oracle-xe-11g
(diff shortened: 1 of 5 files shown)


WellingR:
Note, for some reason our oracle docker image is gone, https://hub.docker.com/r/alexeiled/docker-oracle-xe-11g/ gives a 404 too. I need to investigate why this is...

WellingR:
Fixed, It appears a downgrade of h2 did the trick

WellingR:
Travid did 5 successful builds in a row (for both scala 2.11 and 2.12) so I am confident that the issue is resolved @renatocaval can you do a review?
If you merge, you can do so without squashing, since each commit serves its own purpose

PR #218, opened 2019-01-22
Fix snapshot lookup queries and improve naming for consistency and clarity

richbirch:
We recently found that recovering an actor with large sized snapshots failed due to a missing limit on the query. The query as is stands returns every snapshot meeting the criteria which for us was several 10s of GBs of data.

There was an existing limit on _selectByPersistenceIdAndMaxSeqNr, but none on the other 3 similar queries where a max sequence number or timestamp are specified:

_selectByPersistenceIdAndMaxTimestamp
_selectByPersistenceIdAndMaxSequenceNr
_selectByPersistenceIdAndMaxSequenceNrAndMaxTimestamp

Since Ruud highlighted that the altered queries were also being used for deletions I've added selectOneBy versions of the existing queries to be used only by the snapshot lookups.

I've also renamed most of the query functions to have a consistence convention and make it clearer what their job is

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/snapshot/dao/SnapshotQueries.scala
+++ b/src/main/scala/akka/persistence/jdbc/snapshot/dao/SnapshotQueries.scala
@@ -32,25 +32,25 @@ class SnapshotQueries(val profile: JdbcProfile, override val snapshotTableCfg: S
     SnapshotTable.filter(_.persistenceId === persistenceId).sortBy(_.sequenceNumber.desc)
   val selectAll = Compiled(_selectAll _)
 
-  private def _selectByPersistenceIdAndMaxSeqNr(persistenceId: Rep[String]) =
+  private def _selectLatestByPersistenceId(persistenceId: Rep[String]) =
     _selectAll(persistenceId).take(1)
-  val selectByPersistenceIdAndMaxSeqNr = Compiled(_selectByPersistenceIdAndMaxSeqNr _)
+  val selectLatestByPersistenceId = Compiled(_selectLatestByPersistenceId _)
 
-  private def _selectByPersistenceIdAndSeqNr(persistenceId: Rep[String], sequenceNr: Rep[Long]) =
+  private def _selectByPersistenceIdAndSequenceNr(persistenceId: Rep[String], sequenceNr: Rep[Long]) =
     _selectAll(persistenceId).filter(_.sequenceNumber === sequenceNr)
-  val selectByPersistenceIdAndSeqNr = Compiled(_selectByPersistenceIdAndSeqNr _)
+  val selectByPersistenceIdAndSequenceNr = Compiled(_selectByPersistenceIdAndSequenceNr _)
 
-  private def _selectByPersistenceIdAndMaxTimestamp(persistenceId: Rep[String], maxTimestamp: Rep[Long]) =
+  private def _selectByPersistenceIdUpToMaxTimestamp(persistenceId: Rep[String], maxTimestamp: Rep[Long]) =
     _selectAll(persistenceId).filter(_.created <= maxTimestamp)
-  val selectByPersistenceIdAndMaxTimestamp = Compiled(_selectByPersistenceIdAndMaxTimestamp _)
+  val selectByPersistenceIdUpToMaxTimestamp = Compiled(_selectByPersistenceIdUpToMaxTimestamp _)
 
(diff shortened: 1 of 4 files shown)


richbirch:
Hmm looks like alexeiled/docker-oracle-xe-11g has been pulled from dockerhub

renatocaval:
On that other PR (https://github.com/dnvriend/akka-persistence-jdbc/pull/217/files#diff-354f30a63fb0907d4ad57269548329e3R8), @WellingR also changed the oracle image. 

I believe this supersedes this PR, correct?

renatocaval:
@richbirch, now I see that you are PR is bringing more than just updating the oracle image. 
Can you update this PR? We can keep the oracle image that @WellingR added.

renatocaval:
@richbirch, I fixed the travis conflict and left it as defined by https://github.com/dnvriend/akka-persistence-jdbc/pull/217

richbirch:
@renatocaval i've updated the oracle docker image as requested

richbirch:
And yes, this PR supersedes the earlier one. I had created that one by editing directly in github, so it wasn't easily updatable when I wanted to touch more files

src/main/scala/akka/persistence/jdbc/snapshot/dao/SnapshotQueries.scala
```diff
@@ -32,23 +32,35 @@ class SnapshotQueries(val profile: JdbcProfile, override val snapshotTableCfg: S
...
+  private def _selectOneByPersistenceIdAndMaxTimestamp(persistenceId: Rep[String], maxTimestamp: Rep[Long]) =
+    _selectAll(persistenceId).filter(_.created <= maxTimestamp).take(1)
+  val selectOneByPersistenceIdAndMaxTimestamp = Compiled(_selectOneByPersistenceIdAndMaxTimestamp _)
+
+  private def _selectOneByPersistenceIdAndMaxSequenceNr(persistenceId: Rep[String], maxSequenceNr: Rep[Long]) =
+    _selectAll(persistenceId).filter(_.sequenceNumber <= maxSequenceNr).take(1)
+  val selectOneByPersistenceIdAndMaxSequenceNr = Compiled(_selectOneByPersistenceIdAndMaxSequenceNr _)
+
+  private def _selectOneByPersistenceIdAndMaxSequenceNrAndMaxTimestamp(persistenceId: Rep[String], maxSequenceNr: Rep[Long], maxTimestamp: Rep[Long]) =
+    _selectOneByPersistenceIdAndMaxSequenceNr(persistenceId, maxSequenceNr).filter(_.created <= maxTimestamp).take(1)
```
WellingR: In this query you re-use a selectOne query. I dumped the resulting query and the result is

```sql
select x2, x3, x4, x5 from (select "persistence_id" as x2, "sequence_number" as x3, "created" as x4, "snapshot" as x5 from "snapshot" where ("persistence_id" = ?) and ("sequence_number" <= ?) order by "sequence_number" desc limit 1) where x4 <= ? limit 1
```

the `select ... from (select ... from ... limit 1) where ... limit 1` construction might cause issues, since the "limit" in the subquery can cause a result which should have been returned to be filtered out.

All other queries look fine (and they do not use subqueries at all, so that is good).

WellingR:
@richbirch one more thing:

Dennis Vriend has agreed with Lightbend and the Akka team to hand over the maintenance responsibility of this Akka Persistence plugin. We will do that and the new OSS project will be located at [akka/akka-persistence-jdbc](https://github.com/akka/akka-persistence-jdbc).

All Lightbend projects require that the contributors have agreed to the Lightbend CLA. This is the same CLA as is used for the Akka project. Would you mind signing the CLA online at https://www.lightbend.com/contribute/cla?

The CLA signing is a simple online process, no papers involved. You sign it digitally by providing your email address and GitHub user. That’s all.

Thanks in advance.

src/main/scala/akka/persistence/jdbc/snapshot/dao/SnapshotQueries.scala
richbirch: Excellent catch! I've updated to use the 'UpTo' version of the query for the subquery. I *think* this is ok. Please shout again if I've missed anything

richbirch:
@WellingR Lightbend CLA signed :)

src/main/scala/akka/persistence/jdbc/snapshot/dao/SnapshotQueries.scala
richbirch: I've also rebased / squashed

PR #220, opened 2019-01-28
Dependency updates

WellingR:
Updated scala to 2.12.8 and akka to 2.5.19

### DIFF
--- a/project/ProjectAutoPlugin.scala
+++ b/project/ProjectAutoPlugin.scala
@@ -9,8 +9,8 @@ import SbtScalariform.autoImport._
 import scalariform.formatter.preferences.FormattingPreferences
 
 object ProjectAutoPlugin extends AutoPlugin {
-  final val ScalaVersion = "2.12.7"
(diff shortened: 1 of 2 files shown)


project/ProjectAutoPlugin.scala
```diff
@@ -9,8 +9,8 @@ import SbtScalariform.autoImport._
 import scalariform.formatter.preferences.FormattingPreferences
 
 object ProjectAutoPlugin extends AutoPlugin {
-  final val ScalaVersion = "2.12.7"
-  final val AkkaVersion = "2.5.17"
+  final val ScalaVersion = "2.12.8"
+  final val AkkaVersion = "2.5.19"
```
renatocaval: Akka 2.5.20 is about to be released. Let's wait a few days for this one.

renatocaval: And here it is!

```suggestion
  final val AkkaVersion = "2.5.20"
```

PR #230, opened 2019-05-09
Clean changes in preparation for Scala 2.13

dwijnand:


src/main/scala/akka/persistence/jdbc/util/SlickExtension.scala
```diff
@@ -69,8 +69,8 @@ class DefaultSlickDatabaseProvider(system: ActorSystem) extends SlickDatabasePro
       val conf = confObj.toConfig
       if (conf.hasPath("profile")) {
         // Only create the LazySlickDatabase if a profile has actually been configured, this ensures that the example in the reference conf is ignored
-        Some(key -> new LazySlickDatabase(conf, system))
-      } else None
+        List(key -> new LazySlickDatabase(conf, system))
+      } else Nil
```
ignasi35: Why not use `Map`? IT's the expected type after all, right?

How is using `List/Nil` better than `Some/None`?

dwijnand: > Why not use `Map`? IT's the expected type after all, right?

`Map` would work, but it just has to be something that can be iterated to rebuild the external map.  (`List`/`Nil` is a bit cheaper, but that wasn't a strong consideration.)

> How is using `List/Nil` better than `Some/None`?

In 2.13.0-RC1 Some/None doesn't work, because Map#flatMap is overloaded (which breaks the option2iterable implicit conversion that is required).  If and how to fix this is being discussed, this is one of the use-site fixes.

PR #231, opened 2019-05-09
Cross-build to Scala 2.13.0-RC1

dwijnand:
Builds on #230
Needs a release of Slick for 2.13.0-RC1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ before_install:
 scala:
   - 2.11.12
   - 2.12.8
+  - 2.13.0-RC1
 jdk:
   - oraclejdk8
   - openjdk11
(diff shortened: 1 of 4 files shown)


WellingR:
As long as the new slick version has not yet been released yet, this will not build, so perhaps we should keep this PR open until we are sure we can get it to work.

dwijnand:
Yeah, it's passing because it's not being tested.  I'll push the CI update and then we can wait for Slick.

dwijnand:
That's more like it

```
unresolved dependency: com.typesafe.slick#slick-hikaricp_2.13.0-RC1;3.3.0: not found
```

dwijnand:
Let's pick this one up when Slick moves forward.

PR #234, opened 2019-05-28
Build for Scala 2.13.0-RC2

renatocaval:


renatocaval:
This won't build for now. Slick 3.3.1 is being published to maven central

renatocaval:
@WellingR, this one is green. I would like to cut a release (RC) so we can move on in Lagom.

renatocaval:
Maybe we should cut a 3.5.1 with artefacts for Scala 2.13.0-RC2 instead of 3.6.0-RC1. I can backport it to 3.5.0 if decided so.

renatocaval:
I just realised that there is now 3.5.x branch. We will cut 3.5.1 from master, so no backport needed.

dwijnand:
Who can cut a release of akka-persistence-jdbc for Scala 2.13.0-RC2?  @renatocaval can you?

PR #235, opened 2019-06-03
Enable mima to check binary compatibility

WellingR:
If I understood correctly the release we want to make this will will be version 3.5.1.

I have added and enabled MiMa to check that this release is compatible with the 3.5.0 release.

There is one incompatibility (for scala 2.11 only), which occurs because the type of a lambda changed. As this lambda is not accessible when compiling, this should not be a problem.


renatocaval:
@WellingR, there was an import missing. I just push it. 

We can merge when green.

PR #237, opened 2019-06-03
[WIP] Enable release publishing from Travis

renatocaval:
This PR make it possible to release by creating a tag on GitHub's interface. Once the tag is in place, Travis will build and publish to Bintray.

TODO: 
@WellingR or @dnvriend, there is one extra bit to do to get it working.
One of you need to add two Bintray environment variables to Travis. We need `BINTRAY_PASS` and `BINTRAY_USER` configured in Travis as env variables for the build. 
Once that in place, the Bintray plugin will be able to read it and publish to Bintray. 

The whole process works as follow:

* a maintainer creates a tag using GitHub's release page
* travis 'sees' it and starts a build for publishing
* dynver plugin reads the tag and set current version to the tag, for instance, if tag is v3.5.1, dynver will set project version to `3.5.1`.
* Bintray plugin reads credentials from env var and publish to Bintray
* we still need the manual sync from Bintray to Maven (done by Dennis)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,17 @@ before_script:
   - psql -c 'create database docker;' -U postgres
 addons:
   postgresql: "9.4"
+
 script: "./build.sh"
+
+jobs:
+  include:
+    - stage: publish
+      script: "./sbt-publish.sh"
+
+stages:
+  - name: publish
+    if: repo = dnvriend/akka-persistence-jdbc AND type != pull_request AND tag =~ ^v
 
 before_cache:
   - find $HOME/.ivy2 -name "ivydata-*.properties" -delete
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -28,3 +28,5 @@ addSbtPlugin("de.heikoseeberger" % "sbt-header" % "4.1.0")
 addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.7")
 
 addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.3.0")
+
+addSbtPlugin("com.dwijnand" % "sbt-dynver" % "3.3.0")


WellingR:
Looks good, however I do not have the bintray credentials you need.
What we perhaps could do is create a new bintray account. As syncing to maven central requires a separate login to sonatype.

WellingR:
We could investigate if we can give other maintainers permissions to sync to maven central. As far as I know, this can be done from any bintray repository, you only need to request sonatype access for several users, see https://central.sonatype.org/pages/ossrh-guide.html . In short each user which wants to publish should create an sonatype JIRA account, Dennis (the owner of the repository) can request acces for the jira usernames, and then each user can use his own sonatype jira credentials to publish artifacts (or perhaps we can create an account for travis).

ignasi35:
@renatocaval if we enable `sbt-dynver` (hell, yeah! :-) ), shouldn't this PR also remove `version.sbt`?

renatocaval:
@ignasi35, you are right. And that was the case. The file was re-added when I merged master onto this branch. I'll revert and do a rebase instead.

@WellingR The problem with Sonatype credentials is that @dnvriend will have to give us access to the whole group id `com.github.dnvriend`. On him to decide of course, but I understand that his not optimal.

Instead, we could do like Slick. Slick publishes it directly on Sonatype without bintray using encrypted credentials from Stefan Zeiger. We do it the same here. 

However, in Slick we only do staging and someone with access to `com.typesafe` needs to promote the artifacts. For the plugin, we could stage and promote to avoid that extra step. That would free Dennis from the manual step. 

One other option is to move the project to its own org, say `akka-persistence-jdbc` instead of Akka org and choose another groupId for Sonatype. But that's more work and has the risk of people adding the artefacts twice on the classpath.

renatocaval:
I marked this one as WIP. 

I won't have the time right now, but want to try to get the configuration in play to publish directly to Sonatype. Then we will need @dnvriend  to add the encrypted keys to it.

I think this would be the most desired situation. Release from Travis, directly to Sonatype using tagging.

renatocaval:
I'm closing that one. This is out-of-date now. Part of it has been done in #258. 
The rest will follow.

PR #241, opened 2019-07-24
Only fetch `batch` number of rows at a time from the db

richbirch:
Rather than fetching potentially all the data for a given persistenceId and then only using the first `maxBufferSize` rows, we should only fetch `maxBufferSize` rows in the first place.

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
+++ b/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
@@ -142,8 +142,7 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
         case _ =>
           delaySource
             .flatMapConcat { _ =>
-              currentJournalEventsByPersistenceId(persistenceId, from, toSequenceNr)
-                .take(readJournalConfig.maxBufferSize)
+              currentJournalEventsByPersistenceId(persistenceId, from, from + (readJournalConfig.maxBufferSize - 1))
             }
             .mapConcat(adaptEvents)
             .map(repr => EventEnvelope(Sequence(repr.sequenceNr), repr.persistenceId, repr.sequenceNr, repr.payload))
--- a/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
+++ b/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
@@ -115,7 +115,7 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
           knownIds += id
           xs
         }
-        (id) => next(id)
+        id => next(id)
       }
 
   private def adaptEvents(repr: PersistentRepr): Seq[PersistentRepr] = {
@@ -133,7 +133,7 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
       .map(repr => EventEnvelope(Sequence(repr.sequenceNr), repr.persistenceId, repr.sequenceNr, repr.payload))
 
   override def eventsByPersistenceId(persistenceId: String, fromSequenceNr: Long, toSequenceNr: Long): Source[EventEnvelope, NotUsed] =
-    Source.unfoldAsync[Long, Seq[EventEnvelope]](Math.max(1, fromSequenceNr)) { (from: Long) =>
+    Source.unfoldAsync[Long, Seq[EventEnvelope]](Math.max(1, fromSequenceNr)) { from: Long =>
       def nextFromSeqNr(xs: Seq[EventEnvelope]): Long = {
         if (xs.isEmpty) from else xs.map(_.sequenceNr).max + 1
       }
@@ -142,7 +142,8 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
         case _ =>
           delaySource
             .flatMapConcat { _ =>
-              currentJournalEventsByPersistenceId(persistenceId, from, from + (readJournalConfig.maxBufferSize - 1))
+              val to = Math.min(from + (readJournalConfig.maxBufferSize - 1), toSequenceNr)
+              currentJournalEventsByPersistenceId(persistenceId, from, to)
             }
             .mapConcat(adaptEvents)
             .map(repr => EventEnvelope(Sequence(repr.sequenceNr), repr.persistenceId, repr.sequenceNr, repr.payload))


src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -142,8 +142,7 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
         case _ =>
           delaySource
             .flatMapConcat { _ =>
-              currentJournalEventsByPersistenceId(persistenceId, from, toSequenceNr)
-                .take(readJournalConfig.maxBufferSize)
+              currentJournalEventsByPersistenceId(persistenceId, from, from + (readJournalConfig.maxBufferSize - 1))
```
WellingR: what if `toSequenceNr` is smaller than `from + (readJournalConfig.maxBufferSize - 1)`

Perhaps we could re-use some of the code used for the eventsByTag query. Where a similar mechanism is being used

WellingR:
This is an improvement that has been on my list for some time, but I sort of forgot about improving this. Thanks!

The tests currently fail though...

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
richbirch: Good point. I've just updated with a fix for this case

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -142,8 +142,8 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
         case _ =>
           delaySource
             .flatMapConcat { _ =>
-              currentJournalEventsByPersistenceId(persistenceId, from, toSequenceNr)
-                .take(readJournalConfig.maxBufferSize)
+              val to = Math.min(from + (readJournalConfig.maxBufferSize - 1), toSequenceNr)
```
WellingR: Since it is possible to physically delete events. It could be possible that a persistent actor has its first event with the lowest sequence number 101 (in this case the first 100 events have been deleted).

If we compute the to value this way, and we want to query all events (from 1 to Long.MaxValue), then our first batch would be empty.
It seems that the stream would still continue, so eventually the remaining data would be included int the results.
However since the `delaySource` will keep on triggering delays, each batch will have a delay between them.

WellingR:
I consider this implementation an improvement over the previous implementation, and I would be willing to merge this. However I do think that a solution similar to the eventsByTag query could be used to further improve this implementation.
The main issue is that this implementation (and the previous one) will introduce a delay for each batch of events that is being retrieved, this delay is not needed in case a full batch (of `maxBufferSize` rows) is retrieved.

richbirch:
Thanks for explaining the issue. It sounds like there's some room to
refactor into some common functions. I hadn't appreciated what the code in
the eventsByTag was trying to do in the brief time I had to look at it.

I'm afraid my main concern was to make it useable. It almost immediately
crashes our app in its current form due to fetching more data than can fit
in memory. I only started playing with eventsByPersistenceId after reading
this post on distributed streams in akka:

https://blog.colinbreck.com/integrating-akka-streams-and-akka-actors-part-iv/

Quite an inspiring read.

I'm very grateful to have been able to use this library at all, and I'm
glad to have been able to fix a couple of things on the way. If I can
manage to squeeze in some time to refactor this fix to be more complete
then I'd love to do so. I just can't promise it'll happen any time soon!

WellingR:
I'll try to make sure that either this change will be included in the next release, or that an improved version (like the eventsByTag variant) is included in the next release. (no release date has been set yet)

WellingR:
@renatocaval can you have a look at this and decide if we want to include these changes in our next release?

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
renatocaval: Yes indeed. Do you think we should optimize it and first fetch the lowest non-deleted event seqNum for the passed persistence id? 

On the other hand, someone using this method is really controlling the environment by explicitly passing the `from` and `to` sequence numbers. So, I would not expect people calling it with a `fromSeqNr` for a delete event.

renatocaval: quoting myself...
> So, I would not expect people calling it with a fromSeqNr for a delete event.

And that's when people surprises you. We never know how it will be used.

renatocaval: @richbirch / @WellingR, I'm ok with the currenct implementation. Maybe we should add in the scaladoc that query will be delayed if you have deleted events and you querying using a `fromSequeceNr` from a delete event.

renatocaval: And also a notice on the README section covering `eventsByPersistenceId`
https://github.com/dnvriend/akka-persistence-jdbc#eventsbypersistenceidquery-and-currenteventsbypersistenceidquery

renatocaval:
> or that an improved version (like the eventsByTag variant) 

Do you mean that we should use the same kind of delay mechanism we have in `eventsByTag`? (PR https://github.com/dnvriend/akka-persistence-jdbc/pull/111)

If I recall correctly, the gaps that can occur that we fixed in `eventsByTag` can only happen when you have many interleaved threads adding events. That can only happen when adding events from different persistent ids. 

When querying by `persistenceId` that can't occur because by design there is only one writer. So the queries will always be consistent.

WellingR:
> Do you mean that we should use the same kind of delay mechanism we have in eventsByTag? (PR #111)

Nope, I meant that this implementation could be improved by using the control mechanism from the eventsByTag query, see
https://github.com/dnvriend/akka-persistence-jdbc/blob/ab540559e2100b5e68a2f56c65e7e378adbbfcbc/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala#L207

Specifically, the code could dynamically decide whether to query the next batch with or without delay depending on whether the current batch was a complete batch or not

renatocaval:
@richbirch, thanks for your time in bringing this. Sorry for not merging this PR. Your fix is now covered by #248 with the opitmization that @WellingR was mentioning.

richbirch:
No problem at all. I'm pleased it's lead to a better fix :)

PR #247, opened 2019-10-18
Bump akka version to 2.6.0-RC1

renatocaval:
In Akka 2.6, we should use the internal materializer instead of creating on each time.  

We need to update the code and remove any internal usage of `Materializer`.

The changes on the materializer are also motivated by this 'bug': 
https://github.com/akka/akka/issues/28037

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,13 +165,13 @@ The plugin supports the following queries:
 
 ```scala
 import akka.actor.ActorSystem
-import akka.stream.{Materializer, ActorMaterializer}
+import akka.stream.Materializer
 import akka.stream.scaladsl.Source
 import akka.persistence.query.PersistenceQuery
 import akka.persistence.jdbc.query.scaladsl.JdbcReadJournal
 
 implicit val system: ActorSystem = ActorSystem()
-implicit val mat: Materializer = ActorMaterializer()(system)
+implicit val mat: Materializer = Materializer.matFromSystem(system)
 val readJournal: JdbcReadJournal = PersistenceQuery(system).readJournalFor[JdbcReadJournal](JdbcReadJournal.Identifier)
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -165,13 +165,13 @@ The plugin supports the following queries:
...
 import akka.actor.ActorSystem
-import akka.stream.{Materializer, ActorMaterializer}
+import akka.stream.Materializer
 import akka.stream.scaladsl.Source
 import akka.persistence.query.PersistenceQuery
 import akka.persistence.jdbc.query.scaladsl.JdbcReadJournal
 
 implicit val system: ActorSystem = ActorSystem()
-implicit val mat: Materializer = ActorMaterializer()(system)
+implicit val mat: Materializer = Materializer.matFromSystem(system)
```
johanandren: Preferrably even skip the implicit mat at all, it should be available just by having the system implicit.

renatocaval:
> This will make the plugin require 2.6 though, vs just supporting it (we'll need to figure out the timeout failure on the Akka side).

Yes, I know. And I would prefer to avoid it. 

Isn't that also the case for `akka-persistence-cassandra`?

README.md
renatocaval: sure, I will revisit all this.

WellingR:
It would be cool if `Materializer.matFromSystem(system)` and `Source.future(...)` would be available on akka 2.5.x

renatocaval:
Yeah, let's wait for this one. My main motivation here was to avoid a bug we hit in the Lagom build, but it's preferable to fix it and don't touch the plugin. 

It's always good if we can avoid a compatibility matrix.

renatocaval:
The bug I was mentioned will be fixed in Akka and for Lagom we have a workaround. 

That said, it can close this PR and keep the plugin compatible with both versions of Akka.

PR #248, opened 2019-10-27
events by persistenceid improvements

WellingR:
This improved solution closes #241

Only the configured batch size of events are fetched each time. This implementation is based on the implementation of the eventsByTag query

### DIFF
--- a/project/ProjectAutoPlugin.scala
+++ b/project/ProjectAutoPlugin.scala
@@ -106,8 +106,8 @@ object ProjectAutoPlugin extends AutoPlugin {
  )
 
   def determineMimaPreviousArtifacts(scalaBinVersion: String): Set[ModuleID] = {
-    val compatVersions: Set[String] = if (scalaBinVersion.startsWith("2.13")) Set.empty else {
-      Set("3.5.0")
+    val compatVersions: Set[String] = if (scalaBinVersion.startsWith("2.13")) Set("3.5.2") else {
+      Set("3.5.0", "3.5.1", "3.5.2")
     }
     compatVersions.map { v =>
       "com.github.dnvriend" % ("akka-persistence-jdbc_" + scalaBinVersion) % v
--- a/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
+++ b/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
@@ -128,11 +128,12 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
       .mapAsync(1)(deserializedRepr => Future.fromTry(deserializedRepr))
 
   override def currentEventsByPersistenceId(persistenceId: String, fromSequenceNr: Long, toSequenceNr: Long): Source[EventEnvelope, NotUsed] =
-    currentJournalEventsByPersistenceId(persistenceId, fromSequenceNr, toSequenceNr, Long.MaxValue)
-      .mapConcat(adaptEvents)
-      .map(repr => EventEnvelope(Sequence(repr.sequenceNr), repr.persistenceId, repr.sequenceNr, repr.payload))
+    eventsByPersistenceIdSource(persistenceId, fromSequenceNr, toSequenceNr, currentEventsOnly = true)
 
-  override def eventsByPersistenceId(persistenceId: String, fromSequenceNr: Long, toSequenceNr: Long): Source[EventEnvelope, NotUsed] = {
+  override def eventsByPersistenceId(persistenceId: String, fromSequenceNr: Long, toSequenceNr: Long): Source[EventEnvelope, NotUsed] =
+    eventsByPersistenceIdSource(persistenceId, fromSequenceNr, toSequenceNr, currentEventsOnly = false)
+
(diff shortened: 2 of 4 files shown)


WellingR:
@renatocaval @richbirch please review

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -123,36 +123,53 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
...
+      case (from, control) =>
+        def retrieveNextBatch(): Future[Option[((Long, FlowControl), Seq[PersistentRepr])]] = {
+          for {
+            xs <- currentJournalEventsByPersistenceId(persistenceId, from, toSequenceNr, batchSize).runWith(Sink.seq)
+          } yield {
+            val hasMoreEvents = xs.size == batchSize
+            // Events are ordered by sequence number, therefore the last one is the largest
+            val hasLastEvent = xs.lastOption.exists(last => last.sequenceNr >= toSequenceNr)
+            val nextControl: FlowControl =
+              if (hasLastEvent || from > toSequenceNr) Stop
```
renatocaval: If `hasLastEvent` is true, we should `ContinueDelayed`. This is expected to be an infinite stream of events. 

I think we will need another private method that we can use here in which we indicate if we want to stop on the last event or if we want to continue polling. I mean, `currentEventsByPersistenceId` should also benefit from this optimization. 

For `eventsByTag` we have `private def eventsByTag(tag: String, offset: Long, terminateAfterOffset: Option[Long])`. And we use `terminateAfterOffset` to decide if it should keep polling or not.

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -123,36 +123,53 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
...
+
+    Source.unfoldAsync[(Long, FlowControl), Seq[PersistentRepr]]((Math.max(1, fromSequenceNr), Continue)) {
+      case (from, control) =>
+        def retrieveNextBatch(): Future[Option[((Long, FlowControl), Seq[PersistentRepr])]] = {
+          for {
+            xs <- currentJournalEventsByPersistenceId(persistenceId, from, toSequenceNr, batchSize).runWith(Sink.seq)
+          } yield {
+            val hasMoreEvents = xs.size == batchSize
+            // Events are ordered by sequence number, therefore the last one is the largest
+            val hasLastEvent = xs.lastOption.exists(last => last.sequenceNr >= toSequenceNr)
```
renatocaval: We could have an optimization here: 

```scala
val lastSeqNrInBatch: Option[PersistentRepr] = xs.lastOption.map(_.sequenceNr)
val hasLastEvent = lastSeqNrInBatch.exists(_ >= toSequenceNr)

... then later

val nextFrom: Long = 
  lastSeqNrInBatch match {
    // Continue querying from the last sequence number (the events are ordered)
    case Some(lastSeqNr) => lastSeqNr + 1
    case None =>  from
  }
```

As it is now, we are traversing `xs` twice. It's a tiny optimization, but helpful when `xs` is large.

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
WellingR: The trait we are implementing see [EventsByPersistenceIdQuery.eventsByPersistenceId](https://doc.akka.io/api/akka/current/akka/persistence/query/scaladsl/EventsByPersistenceIdQuery.html#eventsByPersistenceId(persistenceId:String,fromSequenceNr:Long,toSequenceNr:Long):akka.stream.scaladsl.Source[akka.persistence.query.EventEnvelope,akka.NotUsed]) has the `toSequenceNr` as a mandatory parameter, it documents the following:
> You can retrieve a subset of all events by specifying fromSequenceNr and toSequenceNr or use 0L and Long.MaxValue respectively to retrieve all events.

I agree with the point that `currentEventsByPersistenceId` could also benefit from this optimization, this requires retrieving the current `toSequenceNr` first though.

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
WellingR: Implemented this.
I am not sure how much is really matters since `xs` should be a vector, and we only look at the last element, but anyways this is could be slightly better.

project/ProjectAutoPlugin.scala
```diff
@@ -106,8 +106,8 @@ object ProjectAutoPlugin extends AutoPlugin {
  )
 
   def determineMimaPreviousArtifacts(scalaBinVersion: String): Set[ModuleID] = {
-    val compatVersions: Set[String] = if (scalaBinVersion.startsWith("2.13")) Set.empty else {
-      Set("3.5.0")
+    val compatVersions: Set[String] = if (scalaBinVersion.startsWith("2.13")) Set("3.5.2") else {
+      Set("3.5.0", "3.5.1", "3.5.2")
```
renatocaval: We can simplify it by only checking against 3.5.2. Since 3.5.0 and 3.5.1 are binary compatible with 3.5.2. 

This can be on another PR

PR #251, opened 2019-11-18
Changed deleted field to not null

### LINKED ISSUE #250: Why deleted field is nullable?
In all schemas `journal.deleted` field is nullable, however, in Slick it is not:
[Postgres](https://github.com/dnvriend/akka-persistence-jdbc/blob/master/src/test/resources/schema/postgres/postgres-schema.sql#L7):
```sql
CREATE TABLE IF NOT EXISTS public.journal (
----
  deleted BOOLEAN DEFAULT FALSE,
----
);
```

[In Slick:](https://github.com/dnvriend/akka-persistence-jdbc/blob/master/src/main/scala/akka/persistence/jdbc/journal/dao/JournalTables.scala#L36)
```scala
val deleted: Rep[Boolean] = column[Boolean](journalTableCfg.columnNames.deleted, O.Default(false))
```

Was it done intentionally, or should I create PR for that?

feoktant:
According to [discussion here](https://github.com/dnvriend/akka-persistence-jdbc/issues/250), added `NOT NULL` constraint in test/example scripts

Merging this closes #250

### DIFF
--- a/src/test/resources/schema/h2/h2-schema.sql
+++ b/src/test/resources/schema/h2/h2-schema.sql
@@ -4,7 +4,7 @@ CREATE TABLE IF NOT EXISTS PUBLIC."journal" (
   "ordering" BIGINT AUTO_INCREMENT,
   "persistence_id" VARCHAR(255) NOT NULL,
   "sequence_number" BIGINT NOT NULL,
-  "deleted" BOOLEAN DEFAULT FALSE,
+  "deleted" BOOLEAN DEFAULT FALSE NOT NULL,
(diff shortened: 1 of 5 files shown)


WellingR:
Thanks for this contribution, since Lightbend has shown interest to move this module to the official akka repositories, contributors need to sign the CLA.

Would you mind signing the Lightbend CLA at http://www.lightbend.com/contribute/cla ?

feoktant:
I've signed it

ennru:
I checked the CLA manually.

PR #264, opened 2019-12-05
Use scalafmt instead of scalariform

marcospereira:
## Purpose

This removes Scalaliform in favor of scalafmt. It is using a copy of Akka's `.scalafmt.conf` to maintain consistency with other projects in Akka org.

There are multiple commits to make it easier to review.

## Changes

- [x] bye, scalariform
- [x] hello, scalafmt

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,0 +1,52 @@
+version = 2.2.2
+
+style = defaultWithAlign
+
+docstrings                 = JavaDoc
+indentOperator             = spray
+maxColumn                  = 120
+rewrite.rules              = [RedundantParens, SortImports, AvoidInfix]
+unindentTopLevelOperators  = true
+align.tokens               = [{code = "=>", owner = "Case"}]
+align.openParenDefnSite    = false
+align.openParenCallSite    = false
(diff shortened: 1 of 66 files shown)


.scalafmt.conf
```diff
@@ -0,0 +1,52 @@
+version = 2.2.2
```
marcospereira: Using the latest version. This is the only difference regarding Akka's configuration.

.travis.yml
```diff
@@ -23,6 +23,8 @@ jobs:
     - stage: check
       env: CMD="+Test/compile"
       name: "Compile all code with Scala 2.12 and 2.13"
+    - env: CMD="scalafmtCheckAll scalafmtSbtCheck"
```
marcospereira: @ennru maybe this should be a single job in `check` stage. Although I think that having multiple jobs made it clear what is failing.

marcospereira: By the way, just noticed that we are not checking file headers yet (should be another PR though).

marcospereira: By the way, I just noticed that we are not checking file headers yet (should be another PR though).

PR #265, opened 2019-12-05
Scala 2.12.10, 2.13.1; Only install the used JDK

marcospereira:
## Purpose

Avoid installing a version of the JDK that won't be used by the job.

## Changes

- [x] only install the JDK used by the job
- [x] opportunistically update Scala to version 2.12.10
- [x] opportunistically align Scala 2.13 version used by the build and Travis


ennru:
I moved the `jabba install` to the script, otherwise, the version is not known.
(Normally both JDK versions will end up in the cache anyway.)

PR #269, opened 2019-12-06
Build: check against Whitesource in master

ennru:


ennru:
Thank you for reacting, I've now tried it.

PR #271, opened 2019-12-06
 Apply and check copyright header; JPMS name 

ennru:
## Purpose

Apply the new Copyright headers. (commit 2)
Add JPMS module name, enable header plugin and check on Travis. (commit 1)

### DIFF
--- a/project/AutomaticModuleName.scala
+++ b/project/AutomaticModuleName.scala
@@ -15,6 +15,6 @@ import sbt.{ Def, _ }
 object AutomaticModuleName {
   private val AutomaticModuleName = "Automatic-Module-Name"
 
-  def settings(name: String): Seq[Def.Setting[Task[Seq[PackageOption]]]] = Seq(
-    packageOptions in (Compile, packageBin) += Package.ManifestAttributes(AutomaticModuleName -> name))
+  def settings(name: String): Seq[Def.Setting[Task[Seq[PackageOption]]]] =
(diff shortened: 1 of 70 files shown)


project/AutomaticModuleName.scala
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright (C) 2009-2018 Lightbend Inc. <http://www.lightbend.com>
```
renatocaval:
```suggestion
 * Copyright (C) 2009-2019 Lightbend Inc. <http://www.lightbend.com>
```

Will this file also be updated by the hearder plugin? I'm wondering how it will behave with new files that do not have

src/main/scala/akka/persistence/jdbc/query/JournalSequenceActor.scala
```diff
@@ -1,3 +1,8 @@
+/*
+ * Copyright (C) 2014 - 2019 Dennis Vriend <https://github.com/dnvriend>
+ * Copyright (C) 2019 - 2019 Lightbend Inc. <https://www.lightbend.com>
+ */
```
renatocaval: I think this is a good example. New files will also get `Copyright (C) 2014 - 2019 Dennis Vriend <https://github.com/dnvriend>`

ennru: All files get the same header.

PR #286, opened 2019-12-09
Use scalafmt-native on Travis

ennru:


ennru:
Seems to work nicely.

.scala-steward.conf
```diff
@@ -7,6 +7,7 @@ updates.ignore = [
   { groupId = "com.typesafe.akka", artifactId = "akka-slf4j" },
   { groupId = "com.typesafe.akka", artifactId = "akka-persistence-tck" },
   { groupId = "com.typesafe.akka", artifactId = "akka-stream-testkit" },
-  { groupId = "com.typesafe.akka", artifactId = "akka-testkit" }
+  { groupId = "com.typesafe.akka", artifactId = "akka-testkit" },
+  { groupId = "org.scalameta", artifactId = "scalafmt-core" }
```
renatocaval: why you don't want to update scalafmt?

renatocaval:
@ennru, I would prefer to have a job that does all the preliminary checks in one shot. I see that we have a job for the headers and one for the formatting. If we have one job that does both we won't be paying the extra price of starting an sbt process once again. 

The drawback of using the native binary is that we will need to keep in sync the plugin version and the native one. Which probably explain why you excluded it from scala-steward. 

Having one validation for headers and formatting is not a big deal and will simplify things for us with respect to maintenance.

ennru:
This was more of an experiment. Collapsing the verification steps might make it more speedy than this.
I don't think upgrading scalafmt to all-the-latest does add any value at all.

renatocaval:
> I don't think upgrading scalafmt to all-the-latest does add any value at all. 

That's true. I we should maybe review it on Lagom and Play. We get updates every week and it requires re-formatting most of the times. 

Some updates may improve performance, but still, we can do it from time to time and manually.

PR #293, opened 2019-12-10
Docs: use passphrase for rsync

ennru:
## Purpose

* Move docs deploy into a shell script.
* Remove passphrase before rsync from sbt
* Add passphrase as Travis-encrypted value
* Run deploy on all master builds


.travis/travis-to-gustav.sh
```diff
@@ -0,0 +1,11 @@
...
+
+# Decrypt private ssh key file
+# (File encryption and key/IV injection courtesy of Travis: https://docs.travis-ci.com/user/encrypting-files/)
+openssl aes-256-cbc -K $encrypted_9f19d2edcfdd_key -iv $encrypted_9f19d2edcfdd_iv -in .travis/travis_apjdbc_rsa.enc -out /tmp/id_rsa -d
+eval "$(ssh-agent -s)"
+chmod 600 /tmp/id_rsa
+# remove passphrase from private key
+ssh-keygen -p -P "$RSYNC_PASSPHRASE" -N "" -f /tmp/id_rsa
+ssh-add /tmp/id_rsa
+sbt -jvm-opts .jvmopts-travis docs/publishRsync
```
ignasi35: Shouldn't these commands include a way to exit the script when any of the commands fails?

The original one-liner was `foo && bar && baz`, so it was stopping one a commend fails, right?

PS: I'm not a bash specialist

ennru: Everything after will fail anyway...

PR #294, opened 2019-12-10
Docs: update passphrase for rsync

ennru:
* Use an unencrypted private ssh key with passphrase
* Use a passphrase that does not trip Travis


.travis/travis-to-gustav.sh
```diff
@@ -1,11 +1,9 @@
...
-# Decrypt private ssh key file
-# (File encryption and key/IV injection courtesy of Travis: https://docs.travis-ci.com/user/encrypting-files/)
-openssl aes-256-cbc -K $encrypted_9f19d2edcfdd_key -iv $encrypted_9f19d2edcfdd_iv -in .travis/travis_apjdbc_rsa.enc -out /tmp/id_rsa -d
 eval "$(ssh-agent -s)"
-chmod 600 /tmp/id_rsa
 # remove passphrase from private key
+cp .travis/deploy-key /tmp/id_rsa
+chmod 600 /tmp/id_rsa
 ssh-keygen -p -P "$RSYNC_PASSPHRASE" -N "" -f /tmp/id_rsa
 ssh-add /tmp/id_rsa
```
raboof: Interesting approach. I've also used putting the key into the environment entirely as described at https://github.com/olafurpg/sbt-ci-release#travis , but if you ever try that then note that that only works for keys that are not too long :D

raboof: (I guess we have checked that this doesn't echo the `RSYNC_PASSPHRASE` to the build logs before, right?)

ennru: There is `travis encrypt-file` to encrypt the whole thing eg. when not using a passphrase.

PR #298, opened 2019-12-11
Data/DDL migration tool

ennru:
## Purpose

Create a tool to migrate databases from the current database structure to the one required for upcoming versions.

## References

* Based on [Flyway](https://flywaydb.org)
* Just a simple skeleton which tests an SQL and a Java migration for Postgres


project/Dependencies.scala
```diff
@@ -15,21 +15,32 @@ object Dependencies {
   val SlickVersion = "3.3.2"
   val ScalaTestVersion = "3.0.8"
 
-  val Libraries = Seq(
+  val JdbcDrivers = Seq(
+    "org.postgresql" % "postgresql" % "42.2.9",
+    "com.h2database" % "h2" % "1.4.200",
+    "mysql" % "mysql-connector-java" % "8.0.18",
+    "com.microsoft.sqlserver" % "mssql-jdbc" % "7.4.1.jre8")
```
ignasi35: +1

ennru:
You are right, it uses internal tables. That's why I set a specific name for "our" table in https://github.com/akka/akka-persistence-jdbc/pull/298/files#diff-c18bd911ab5d8d8ff76ea4a4abc24ed5R16

ignasi35:
> You are right, it uses internal tables. That's why I set a specific name for "our" table in [akka/akka-persistence-jdbc/pull/298/files#diff-c18bd911ab5d8d8ff76ea4a4abc24ed5R16](https://github.com/akka/akka-persistence-jdbc/pull/298/files#diff-c18bd911ab5d8d8ff76ea4a4abc24ed5R16)

Oh! I didn't know that feature and overlooked the piece of code. 

This looks really promising! :-)

migration/src/main/resources/db/migration/V002__test-2.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS public.migrated2 (
+  persistence_id VARCHAR(255) NOT NULL,
+  sequence_number BIGINT NOT NULL,
+  created BIGINT NOT NULL,
+  snapshot BYTEA NOT NULL
+);
+CREATE TABLE test_user (
+  name VARCHAR(200)
+);
```
renatocaval: We can move these files to the `test` folder instead. They are useful to test the migration tool itself.

migration/src/main/scala/db/migration/V003__UpdateUsers.scala
```diff
@@ -0,0 +1,20 @@
...
+
+
+  @throws[Exception]
+  override def migrate(context: Context): Unit = {
+    try {
+      val statement = context.getConnection.prepareStatement("INSERT INTO test_user (name) VALUES ('Obelix')")
+      try statement.execute
+      finally if (statement != null) statement.close()
+    }
+  }
```
renatocaval: That's great that we can have a Java/Scala file to do the migration. This will be useful for the unwrapping of the payload and for re-writing the tags. 

Also, when a user wants to re-shard the tags, they can add a class that redefines the sharding number.

migration/src/test/resources/postgres/init.sql
```diff
@@ -0,0 +1,25 @@
...
+
+DROP TABLE IF EXISTS public.snapshot;
+
+CREATE TABLE IF NOT EXISTS public.snapshot (
+  persistence_id VARCHAR(255) NOT NULL,
+  sequence_number BIGINT NOT NULL,
+  created BIGINT NOT NULL,
+  snapshot BYTEA NOT NULL,
+  PRIMARY KEY(persistence_id, sequence_number)
+);
```
renatocaval: I would add this file as the V1 file of a migration. We will need a different set of files for each supported DB. 

Then we can create the table as it is now and apply V2 (creating a manifest column and unwrapping the payload), apply V3 to migrate the tags to a separated table, etc.

project/Dependencies.scala
renatocaval: Hmm, where is the Oracle driver?

migration/src/main/resources/db/migration/V002__test-2.sql
ennru: But these are meant to become the real migrations which we test with the tests. They should be part of what is tested, not of the tests.
We need to split those per supported database.

ennru:
The Oracle driver is in `core/lib` right now. But it is available from Maven Central nowadays.

migration/src/main/resources/db/migration/V002__test-2.sql
renatocaval: Ok, I see you what your intention is now. 

I'm wondering how this will evolve in time. If we add some migration in 4.1.0, a user that started using the plugin in 4.0.0 won't have a migration table and the tool will apply all of them.
If we ship migrations in a jar, we must be sure that they are all idempotent. Easy to do in SQL, but those written in Java will need extra checks.

migration/src/main/scala/db/migration/V003__UpdateUsers.scala
ignasi35: Code-based migrations in Flyway are my all-time favorite feature.

ennru:
> I think we can move the migration files to the test folder and merge it. On another PR we can improve things and explore some ideas

I made the migrations DB vendor specific so that they are independent. I'm not sure what migrations are needed yet, but I suggest we merge this and take it from there.

Issue #270, opened 2019-12-06
Fix compilation with Akka 2.6 (scheduled builds)


ennru:
The use of the Akka Javadsl from Scala seems to trip things up when compiling against Akka 2.6

https://travis-ci.com/akka/akka-persistence-jdbc/jobs/263762979#L401

ennru:
Fixed with #299

PR #299, opened 2019-12-11
Fix compilation for Akka 2.6

ennru:
## Purpose

Make compilation with Akka 2.6 work in the scheduled Travis job.

## Background Context

The use of the Akka 2.6 javadsl from Scala code has some issues with type-inference. This is solved with explicit types.

See #270


ennru:
Yes, we've seen this before in other uses of the javadsl from Scala code.

Issue #300, opened 2019-12-13
Configure serialization to work with Akka 2.6


ennru:
### Short description

The tests fail on the scheduled build with Akka 2.6 as it tries to use Java serialization.

### Details

https://travis-ci.com/akka/akka-persistence-jdbc/jobs/266803245#L447

patriknw:
Fix for JournalPerfSpec.Cmd is in PR https://github.com/akka/akka/pull/28333

patriknw:
Until fixed the workaround is config `akka.actor.allow-java-serialization = on`

ennru:
Solved with #305

PR #305, opened 2019-12-17
Explicitly allow Java serailisation

ennru:
## Purpose

Allow Java serialization which is disallowed by default in Akka 2.6.

## References

#300

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+// Required until https://github.com/akka/akka/pull/28333 is available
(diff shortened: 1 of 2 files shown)


core/src/main/resources/reference.conf
```diff
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+// Required until https://github.com/akka/akka/pull/28333 is available
+akka.actor.allow-java-serialization = on
```
WellingR: Shouldn't each application decide for itself whether to enable this or not?
Do we use java serialization in the tests? If so, then it might make more sense to enable this in the test config

ennru: Ah, you're right, it should just be for our tests.

PR #313, opened 2020-01-08
Update sbt-headers and add missing headers

renatocaval:
This update the sbt-header plugin and add headers on files in the `migration` module. I have no idea how #298 got a green build.


ignasi35:
> I have no idea how #298 got a green build.

That build happened (and succeeded) before #306 was merged. ;-)

renatocaval:
I don't think it's correct, but...
https://github.com/akka/akka-persistence-jdbc/pull/271#discussion_r354817449

renatocaval:
The cron job is failing because of formatting and missing headers. Can we merge this and open an issue to discuss what to do with the headers?

Issue #318, opened 2020-01-10
New schema and migration for 5.0.0


octonato:
Events and Snapshots are wrapped in some Akka classe. The goal is to unwrap because those classes were never intended to serialized and saved to DB.

There is a [branch](https://github.com/akka/akka-persistence-jdbc/tree/akka-serialization) in which this was fixed by James (original PR: [#180]( https://github.com/akka/akka-persistence-jdbc/pull/180)). 

However, we may not integrate it into master and start over again. The main reason is that the PR is done as such that allows users to keep the old column and have mixed mode in which old data is kept around using the wrappers and new data is saved unwrapped. 

This strategy amounts to the complexity of the code, plus the functionality is enabled/disabled by flags. 

Instead, our current goal is to provide a new schema and a migration tool to allow previous journals to be migrated. That means that users moving to 4.0.0 will have to shutdown the system, run the migration and restarting using the new version. 

This is broken down in a three main tasks:

* Journal Store refactoring - #315 
* Snapshot Store refactoring - #316 
* Migration tool - #317

WellingR:
I must say I really like the the forwards/backwards compatibiliy, since it allowed the application to read old events, and write them in the new way. The migration could be executed while the application was running.
My idea was to follow the 4.0.0 release with another release (5.0.0?) which does not have all the compatibility code and flags.

I definitely agree that having all the extra code and complexity for the compatibility is undesirable. However on the other hand an intermediate release might make the migration process a lot more bearable for our users...

renatocaval:
True, it will be much nicer to be forward/backward compatible, but the cost is also high for us. 

Maybe, an alternative approach would be to maintain a branch 3.5.x for a while so users don't need to immediately migrate to 4.0.0. They can decide to do the move on their own pace. 

Concerning the issue related to unwrap the BLOB, this is not a major one and I don't think we need to fix in 3.5.x. Users can keep using it until they are ready for 4.0.0. 

The major issue is the `eventsByTag` query, but that's mainly impacting Lagom users because of how we shard the tags. Of course, it can also impact other users. For instance if you have two different kinds of PersistentActors that tag with `UserCreated` and `AdminUserCreated` each. A possible quick fix for this could be a DAO that doesn't do a `=` select instead of `LIKE`, but has the limitation of one tag per event. 

[...]

PR #329, opened 2020-01-30
Filter tags on the fly - workaround for #168

renatocaval:
As discussed here: https://github.com/akka/akka-persistence-jdbc/issues/318

We plan to cut a release for users that can't immediately migrate to the upcoming 4.0.0 (because of schema migrations). 

This is a quick fix for the most annoying bug we have. When using sharded tags (like in Lagom), the `eventsByTag` call may return events with unrelated tags. For instance, if you search for the tag "foo1", you also get events tag "foo10". 

The workaround is to filter out any partial match, in memory. Once merged and release, this can immediately benefit existing Lagom users.

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -25,6 +25,11 @@ akka-persistence-jdbc {
   #
   logicalDeletion.enable = true
 
+
+  # The tag separator to use when tagging events with more than one tag.
+  # This property affects jdbc-journal.tagSeparator and jdbc-read-journal.tagSeparator.
+  tagSeparator = ","
+
   database-provider-fqcn = "akka.persistence.jdbc.util.DefaultSlickDatabaseProvider"
 
   shared-databases {
@@ -123,7 +128,10 @@ jdbc-journal {
     }
   }
 
-  tagSeparator = ","
+  # The tag separator to use when tagging events with more than one tag.
+  # should not be configured directly, but through property akka-persistence-jdbc.tagSeparator
+  # in order to keep consistent behavior over write/read sides
+  tagSeparator = ${akka-persistence-jdbc.tagSeparator}
 
   dao = "akka.persistence.jdbc.journal.dao.ByteArrayJournalDao"
 
@@ -390,7 +398,10 @@ jdbc-read-journal {
     }
   }
 
-  tagSeparator = ","
+  # The tag separator to use when tagging events with more than one tag.
+  # should not be configured directly, but through property akka-persistence-jdbc.tagSeparator
+  # in order to keep consistent behavior over write/read sides
+  tagSeparator = ${akka-persistence-jdbc.tagSeparator}
 
   slick {
 
--- a/core/src/main/scala/akka/persistence/jdbc/query/dao/ByteArrayReadJournalDao.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/query/dao/ByteArrayReadJournalDao.scala
@@ -10,10 +10,11 @@ import akka.NotUsed
 import akka.persistence.PersistentRepr
 import akka.persistence.jdbc.config.ReadJournalConfig
 import akka.persistence.jdbc.journal.dao.ByteArrayJournalSerializer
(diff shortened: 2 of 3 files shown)


core/src/main/scala/akka/persistence/jdbc/query/dao/ByteArrayReadJournalDao.scala
```diff
@@ -60,6 +63,21 @@ trait BaseByteArrayReadJournalDao extends ReadJournalDao {
...
+   * Retains every event with tags that perfectly match passed tag.
+   * This is a workaround for bug https://github.com/akka/akka-persistence-jdbc/issues/168
+   *
+   * Builds a Source using the passed DatabasePublisher and
+   * filter out all events that where wrongly selected because of the usage ot LIKE operantor.
+   */
+  private[dao] def perfectlyMatchTag(
+      dbPublisher: DatabasePublisher[JournalRow],
+      tag: String): Source[JournalRow, NotUsed] = {
+    Source.fromPublisher(dbPublisher).filter(_.tags.exists(tags => tags.split(",").contains(tag)))
```
renatocaval: I just realised that it's possible to change the tag separator and therefore the split must take that into consideration. I will change this.

core/src/main/scala/akka/persistence/jdbc/query/dao/ByteArrayReadJournalDao.scala
```diff
@@ -60,6 +63,21 @@ trait BaseByteArrayReadJournalDao extends ReadJournalDao {
...
+  /*
+   * Retains every event with tags that perfectly match passed tag.
+   * This is a workaround for bug https://github.com/akka/akka-persistence-jdbc/issues/168
+   *
+   * Builds a Source using the passed DatabasePublisher and
+   * filter out all events that where wrongly selected because of the usage ot LIKE operantor.
+   */
+  private[dao] def perfectlyMatchTag(
+      dbPublisher: DatabasePublisher[JournalRow],
+      tag: String): Source[JournalRow, NotUsed] = {
```
patriknw: doesn't matter, but this is conceptually a `Flow`, and could be defined as such

renatocaval: I can refactor it. I don't have the reflex to think in terms of Flow. Will work on that. ;-)

core/src/main/resources/reference.conf
```diff
@@ -25,6 +25,11 @@ akka-persistence-jdbc {
   #
   logicalDeletion.enable = true
 
+
+  # The tag separator to use when tagging events with more than one tag.
+  # This property affects jdbc-journal.tagSeparator and jdbc-read-journal.tagSeparator.
+  tagSeparator = ","
```
renatocaval: This property was defined in two different places, but we should have only one. We should use the same on the write and read side. 

`jdbc-journal.tagSeparator` and `jdbc-read-journal.tagSeparator` are now referencing this value.

core/src/main/scala/akka/persistence/jdbc/query/dao/ByteArrayReadJournalDao.scala
```diff
@@ -60,6 +67,15 @@ trait BaseByteArrayReadJournalDao extends ReadJournalDao {
...
 }
 
+object TagFilterFlow {
+  /*
+   * Returns a Flow that retains every event with tags that perfectly match passed tag.
+   * This is a workaround for bug https://github.com/akka/akka-persistence-jdbc/issues/168
+   */
+  private[dao] def perfectlyMatchTag(tag: String, separator: String) =
+    Flow[JournalRow].filter(_.tags.exists(tags => tags.split(separator).contains(tag)))
+}
```
renatocaval: @patriknw, refactor to a Flow

renatocaval: It's now taking into consideration the `separator` that is defined in config file.

WellingR:
Looks good

PR #335, opened 2020-02-05
Previous artifact detection

### LINKED ISSUE #285: Detect version of former group ID on classpath
### Short description

Detect if this plugin of the old 3.5.x series is on the classpath and make it fail if so.

### Details

As the group ID changes to "com.lightbend.akka" users may end up with two versions of this plugin on their classpath.
Find a way to detect that and fail.

renatocaval:
Refs #285

Plugin will fail to load if it detects that previous artifact is on the
classpath.

### DIFF
--- a/core/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
@@ -12,7 +12,7 @@ import akka.actor.{ ActorSystem, ExtendedActorSystem }
 import akka.persistence.jdbc.config.JournalConfig
 import akka.persistence.jdbc.journal.JdbcAsyncWriteJournal.{ InPlaceUpdateEvent, WriteFinished }
 import akka.persistence.jdbc.journal.dao.{ JournalDao, JournalDaoWithUpdates }
-import akka.persistence.jdbc.util.{ SlickDatabase, SlickExtension }
+import akka.persistence.jdbc.db.{ SlickDatabase, SlickExtension }
 import akka.persistence.journal.AsyncWriteJournal
 import akka.persistence.{ AtomicWrite, PersistentRepr }
 import akka.serialization.{ Serialization, SerializationExtension }
@@ -25,6 +25,7 @@ import scala.collection.immutable._
 import scala.concurrent.{ ExecutionContext, Future }
 import scala.util.{ Failure, Success, Try }
 import akka.pattern.pipe
+import akka.persistence.jdbc.util.PluginVersionChecker
 
 object JdbcAsyncWriteJournal {
   private case class WriteFinished(pid: String, f: Future[_])
@@ -40,10 +41,13 @@ object JdbcAsyncWriteJournal {
 }
 
 class JdbcAsyncWriteJournal(config: Config) extends AsyncWriteJournal {
+
   implicit val ec: ExecutionContext = context.dispatcher
   implicit val system: ActorSystem = context.system
   implicit val mat: Materializer = ActorMaterializer()
   val journalConfig = new JournalConfig(config)
+
+  PluginVersionChecker.check()
 
   val slickDb: SlickDatabase = SlickExtension(system).database(config)
   def db: Database = slickDb.database
--- a/core/src/main/scala/akka/persistence/jdbc/query/javadsl/JdbcReadJournal.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/query/javadsl/JdbcReadJournal.scala
@@ -10,6 +10,7 @@ import akka.persistence.jdbc.query.scaladsl.{ JdbcReadJournal => ScalaJdbcReadJo
 import akka.persistence.query.{ EventEnvelope, Offset }
 import akka.persistence.query.javadsl._
 import akka.stream.javadsl.Source
+import akka.persistence.jdbc.util.PluginVersionChecker
 
 object JdbcReadJournal {
   final val Identifier = ScalaJdbcReadJournal.Identifier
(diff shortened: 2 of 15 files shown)


renatocaval:
I wrote some more context on the original issue: https://github.com/akka/akka-persistence-jdbc/issues/285#issuecomment-582421513

I realise that there is a potential risk when mixing both versions.

> There is a potential risk here tough. User may never load 3.6.x during development and Murphy law will make sure that 3.6.x gets loaded first in production, the check is run and BOOM!

We may not throw an exception but only add a warning log.

core/src/main/scala/akka/persistence/jdbc/util/PluginVersionChecker.scala
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (C) 2014 - 2019 Dennis Vriend <https://github.com/dnvriend>
+ * Copyright (C) 2019 - 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.persistence.jdbc.util
+
+object PluginVersionChecker {
+  def check() = 
+    try {
+      Class.forName("akka.persistence.jdbc.util.DefaultSlickDatabaseProvider")
```
renatocaval: We check for `akka.persistence.jdbc.util.DefaultSlickDatabaseProvider` that only existing 3.5.0 or prior.

core/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
```diff
@@ -40,11 +41,14 @@ object JdbcAsyncWriteJournal {
 }
 
 class JdbcAsyncWriteJournal(config: Config) extends AsyncWriteJournal {
+  
   implicit val ec: ExecutionContext = context.dispatcher
   implicit val system: ActorSystem = context.system
   implicit val mat: Materializer = ActorMaterializer()
   val journalConfig = new JournalConfig(config)
 
+  PluginVersionChecker.check()
```
renatocaval: The most popular plugin component is the write journal. We don't need to check when loading the snapshot component because it's unlikely that someone will be using the snapshot without the journal. That would even not work.

core/src/main/scala/akka/persistence/jdbc/query/javadsl/JdbcReadJournal.scala
```diff
@@ -23,6 +24,9 @@ class JdbcReadJournal(journal: ScalaJdbcReadJournal)
     with EventsByPersistenceIdQuery
     with CurrentEventsByTagQuery
     with EventsByTagQuery {
+
+  PluginVersionChecker.check()
```
renatocaval: Also checking when loading the read-side component.

It's very common to start queries on application bootstrap. That will force the check in a very early stage.

core/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -54,8 +55,13 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
     with EventsByPersistenceIdQuery
     with CurrentEventsByTagQuery
     with EventsByTagQuery {
+
+  PluginVersionChecker.check()
```
renatocaval: same as above, but for scaladsl

core/src/main/scala/akka/persistence/jdbc/util/PluginVersionChecker.scala
```diff
@@ -0,0 +1,19 @@
...
+
+object PluginVersionChecker {
+  def check() =
+    try {
+      Class.forName("akka.persistence.jdbc.util.DefaultSlickDatabaseProvider")
+      throw new RuntimeException(
+        "Old version of Akka Persistence JDBC found on the classpath. Remove `com.github.dnvriend:akka-persistence-jdbc` from the classpath..")
+    } catch {
+      case _: ClassNotFoundException =>
+      // ALl good! That's intentional.
```
ignasi35:
```suggestion
      // All good! That's intentional.
```

core/src/main/scala/akka/persistence/jdbc/util/PluginVersionChecker.scala
```diff
@@ -0,0 +1,19 @@
...
+package akka.persistence.jdbc.util
+
+object PluginVersionChecker {
+  def check() =
+    try {
+      Class.forName("akka.persistence.jdbc.util.DefaultSlickDatabaseProvider")
+      throw new RuntimeException(
+        "Old version of Akka Persistence JDBC found on the classpath. Remove `com.github.dnvriend:akka-persistence-jdbc` from the classpath..")
+    } catch {
+      case _: ClassNotFoundException =>
```
patriknw: In corresponding Akka utility we also handle `NoClassDefFoundError` but maybe that is not needed here?
https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/actor/ReflectiveDynamicAccess.scala#L51

renatocaval: If I understand correctly, `NoClassDefFoundError` happens when the bytecode references it and it's not found at runtime. So, not applicable here. 

Btw, I didn't think about it. I just got inspiration (copied) from the akka discovery check. There we only catch `ClassNotFoundException`. And probably for the same reasons, we don't depend on it at bytecode level.

core/src/main/scala/akka/persistence/jdbc/query/javadsl/JdbcReadJournal.scala
WellingR: Actually isn't the java dsl using forwarding the calls to the scala dsl? I think checking only for the scala dsl should be sufficient

renatocaval: Ok, thanks for the heads-up. I will check and adapt if necessary.

PR #346, opened 2020-02-07
Include timestamp in PersistentRepr and EventEnvelope

renatocaval:


src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
```diff
@@ -86,8 +86,18 @@ class JdbcAsyncWriteJournal(config: Config) extends AsyncWriteJournal {
...
-    val persistenceId = messages.head.persistenceId
+
+    // add timestamp to all payloads in all AtomicWrite messages
+    val timedMessages =
+      messages.map { atomWrt =>
+        // since they are all persisted atomically,
+        // all PersistentRepr on the same atomic batch gets the same timestamp
+        val now = System.currentTimeMillis()
+        atomWrt.copy(payload = atomWrt.payload.map(pr => pr.withTimestamp(now)))
+      }
```
renatocaval: I was in doubt where to add this. 

Initially, I thought that this should be added by Akka itself when internally creating the `Seq[AtomicWrite]`.

The problem is that the sequence is created using an internal mutable var so each `PersistenceRepr` would get a different timestamp. Not such a big issue though. 

Another reason is, depending on the plugin, this timestamp could differ from the timestamp that is added to the envelope.  For instance, Cassandra plugin will use the Offset, so filling it by default is useless (or even confusing when debugging). 

I guess each plugin implementation should decide when to fill it and how to build the `EventEnvelope`. 

Anyway, adding it here like that for the moment. We can discuss if it's worth or more 'correct' to let Akka fill this field.

src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -171,7 +171,7 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
     }
       .mapConcat(identity)
       .mapConcat(adaptEvents)
-      .map(repr => EventEnvelope(Sequence(repr.sequenceNr), repr.persistenceId, repr.sequenceNr, repr.payload))
+      .map(repr => EventEnvelope(Sequence(repr.sequenceNr), repr.persistenceId, repr.sequenceNr, repr.payload, repr.timestamp))
```
renatocaval: It's missing test for this code path. I will add it.

renatocaval: done

src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
patriknw: `asyncWriteMessages` is the right place.
The reason for not setting it in Akka itself is that journals will anyway store it in different ways and some (like Cassandra) may already have a timestamp. Also, serializing the PersistenceRepr is not the recommended way, so the journal will typically have a separate field for the timestamp (but that is for later in the jdbc plugin).

Issue #322, opened 2020-01-21
Out of memory exceptions under high load


Sh1ftry:
Hello,
we have been using Cassandra plugin and have decided to switch to JDBC plugin with PostgreSQL. However, we have run into an issue, that stops us from migrating. Our service is unable to start, due to out of memory exceptions, when restarting under high load, with many events in journal.
```
postgres=> select count(*) from journal group by persistence_id;
 count
--------
 133894
 205311
 233498
 165092
 166375
 316931
 144082
 191159
 325597
 181759
 173260
 325610
 157724
 233744
 341326
 157182
 214548
 320728
(18 rows)
```

Plugin configuration:
```
slick {
  profile = "slick.jdbc.PostgresProfile$"
  db {
    host = "localhost"
    host = ${?POSTGRES_HOST}
    url = "jdbc:postgresql://"${slick.db.host}":5432/postgres?reWriteBatchedInserts=true"
    user = "postgres"
    user = ${?POSTGRES_USER}
    password = "postgres"
    password = ${?POSTGRES_PASSWORD}
    driver = "org.postgresql.Driver"
    numThreads = 5
    numThreads = ${?POSTGRES_THREADS}
    maxConnections = 5
    maxConnections = ${?POSTGRES_MAX_CONNECTIONS}
    minConnections = 1
    minConnections = ${?POSTGRES_MIN_CONNECTIONS}
  }
}

akka.persistence {
  journal.plugin = "jdbc-journal"
  snapshot-store.plugin = "jdbc-snapshot-store"
  at-least-once-delivery {
    warn-after-number-of-unconfirmed-attempts=3
    max-unconfirmed-messages=1000
    redeliver-interval=10s
    redeliver-interval=${?AT_LEAST_ONCE_DELIVERY_REDELIVER_INTERVAL}
    redelivery-burst-limit=50
  }
}


jdbc-journal {
  slick = ${slick}

  batchSize = 40
  batchSize = ${?POSTGRES_JOURNAL_BATCH_SIZE}
  parallelism = 6
  parallelism = ${?POSTGRES_JOURNAL_PARALLELISM}

  event-adapters {
    msg-dispatched-adapter = "abc.xyz.MessageDispatchedAdapter"
    msg-processed-adapter = "abc.xyz.MessageProcessedAdapter"
  }

  event-adapter-bindings {
    "xyz.Abc$MessageProcessed" = msg-dispatched-adapter
[...]

PR #351, opened 2020-02-13
Set statement parameters to improve memory usage in message query

WellingR:
This is a possible fix for #322
According to https://scala-slick.org/doc/3.3.2/dbio.html#streaming these parameters are needed for postgres to ensure that not all data is retrieved at once (by the jdbc driver) when executing a query.

Todos (help wanted)
- [x] Make the fetch size configurable 
- [ ] Create a test where we can configure the number of events we store, in which we can reproduce the memory issue as described in #322. See if this fix actually improves the situaties

### DIFF
--- a/core/src/test/scala/akka/persistence/jdbc/query/JournalDaoStreamMessagesMemoryTest.scala
+++ b/core/src/test/scala/akka/persistence/jdbc/query/JournalDaoStreamMessagesMemoryTest.scala
@@ -1,0 +1,133 @@
+/*
+ * Copyright (C) 2014 - 2019 Dennis Vriend <https://github.com/dnvriend>
+ * Copyright (C) 2019 - 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.persistence.jdbc.query
+
+import java.util.UUID
+
+import akka.Done
+import akka.actor.ActorSystem
+import akka.persistence.{ AtomicWrite, PersistentRepr }
+import akka.persistence.jdbc.journal.dao.{ ByteArrayJournalDao, JournalTables }
+import akka.serialization.SerializationExtension
+import akka.stream.ActorMaterializer
+import akka.stream.scaladsl.{ Sink, Source }
+import com.typesafe.config.{ ConfigValue, ConfigValueFactory }
+import org.scalatest.concurrent.PatienceConfiguration.Timeout
+import org.slf4j.LoggerFactory
+
+import scala.collection.immutable
+import scala.concurrent.{ Await, ExecutionContextExecutor, Future }
+import scala.concurrent.duration._
+import scala.util.{ Failure, Random, Success }
+
+object JournalDaoStreamMessagesMemoryTest {
+
+  val configOverrides: Map[String, ConfigValue] = Map("jdbc-journal.fetch-size" -> ConfigValueFactory.fromAnyRef("100"))
+}
+
+abstract class JournalDaoStreamMessagesMemoryTest(configFile: String)
+    extends QueryTestSpec(configFile, JournalDaoStreamMessagesMemoryTest.configOverrides)
+    with JournalTables {
+  private val log = LoggerFactory.getLogger(this.getClass)
+
+  val journalSequenceActorConfig = readJournalConfig.journalSequenceRetrievalConfiguration
+  val journalTableCfg = journalConfig.journalTableConfiguration
+
+  import profile.api._
+
+  implicit val askTimeout = 50.millis
+
+  def generateId: Int = 0
+
+  behavior.of("Replaying Persistence Actor")
+
+  it should "stream events" in {
+    withActorSystem { implicit system: ActorSystem =>
+      withDatabase { db =>
+        implicit val mat: ActorMaterializer = ActorMaterializer()
+        implicit val ec: ExecutionContextExecutor = system.dispatcher
+
+        val persistenceId = UUID.randomUUID().toString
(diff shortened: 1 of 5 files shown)


renatocaval:
@WellingR, I made fetchSize configurable. 

I'm off next week. I believe, I hope, to have some more time to add the tests the week after.

renatocaval:
@WellingR, I'm wondering about how to test it. Adding a huge amount of events doesn't give us the guarantees to assess if we fix it or not. Which number is too many? In which environment? 

The ideal test would be to just prove that we are not loading everything. I'm just thinking aloud: if we could insert 30 events where number 30 is a bogus one that can't be read. Then we configure the `fetchSize` to be 10. We should be able to read up to event number 20. The stream will fail when we load events from 21 to 30. That will prove that `fetchSize` is effectivelly loading the stream by chunks. Maybe worth experimenting with that. 

Anyway, thanks for taking care of this.

WellingR:
Creating an automated test for this that works on all environments is very tricky.
What we can do is create a test where the number of events is configuable. On a local environment we can configure the memory settings such that at most 100MB of memory is used. This should be sufficient to validate that we can reproduce this problem with the old version of the code. And hopefully this same test will succeed with this fix. (note: the in-memory h2 database keeps all data in-memory, so testing memory usage is not relevant in this case)

I am actually not sure if all types of databases really support this fetchSize. If this is not the case then we must either choose to have this issue fixed for only those databases which do support it.
Or we must fetch the batches in the same way as is done for the eventsByPersistenceId query.

renatocaval:
My idea of using a bogus event was also to confirm that `fetchSize` works for all DBs. 

But also, to avoid the complexity of tweaking the test per environment. If we can prove that we are not loading everything in one shot for all DBs, we can be confident that we have a good fit.

renatocaval:
I would like to avoid a test based on time and memory usage because that's too flaky. I prefer something that proves that we are batching the fetch operation for all supported DBs. 

It doesn't matter how much we are putting in memory, the important part is to prove that we are not putting everything in memory at the same time. A table with 3 records and a `fetchSize` of 1 is enough if we can find the means to prove that we are not loading 3 rows into memory.

Users can then tweak the fetch size according to their needs.

renatocaval:
My idea of using a bogus payload won't work at all because the deserialization happens much later when everything is already in memory. The issue here is completely opaque from outside slick.

renatocaval:
I've managed to write a test that generates events according to the amount of available memory and then fetches it using stream. 

It was a bit tricky because each DB docker has its own sensibilities. For instance, the Oracle docker has some memory limits so I need to find the exact memory limitation to use. It's possible to tweak that Oracle DB, but I didn't want to take that path. 

The bad news is that MySQL gives an OutOfMemory when using `withStatementParameters`. Seems to have no effect. 

I will do some more clean-up and push what I have so far, but we will have to change the underlying implementation to something similar to `eventsByTag`.

project/ProjectAutoPlugin.scala
```diff
@@ -35,6 +35,7 @@ object ProjectAutoPlugin extends AutoPlugin {
     crossScalaVersions := Dependencies.ScalaVersions,
     scalaVersion := Dependencies.Scala212,
     Test / fork := true,
+    Test / javaOptions ++= Seq("-Xms64m", "-Xmx128m"),
```
renatocaval: This is not optimal. I want to make it different. 

My plan is to exclude the test from the normal run and have another job that runs only this memory test. 

We will need an sbt task that tweaks the `javaOptions` for it. Passing `-Xms64m -Xmx128m` is not good as it is not passed over to forked tests.

core/src/test/scala/akka/persistence/jdbc/query/JournalDaoStreamMessagesMemoryTest.scala
```diff
@@ -0,0 +1,133 @@
...
+}
+
+class PostgresJournalDaoStreamMessagesMemoryTest
+    extends JournalDaoStreamMessagesMemoryTest("postgres-application.conf")
+    with PostgresCleaner
+
+// TODO: MySQL not working with withStatementParameters
+//class MySQLJournalDaoStreamMessagesMemoryTest
+//  extends ReplayPersistenceActorMemoryTest("mysql-application.conf")
+//    with MysqlCleaner
```
renatocaval: Mysql not working at all. We need to revisite the `messages()` impl.

core/src/test/scala/akka/persistence/jdbc/query/JournalDaoStreamMessagesMemoryTest.scala
```diff
@@ -0,0 +1,133 @@
...
+
+class OracleJournalDaoStreamMessagesMemoryTest
+    extends JournalDaoStreamMessagesMemoryTest("oracle-application.conf")
+    with OracleCleaner
+
+class SqlServerJournalDaoStreamMessagesMemoryTest
+    extends JournalDaoStreamMessagesMemoryTest("sqlserver-application.conf")
+    with SqlServerCleaner
```
renatocaval: Oracle and SqlServer works out-of-the-box (also without withStatementParameters). The driver don't keep things in memory.

patriknw:
Superseded by https://github.com/akka/akka-persistence-jdbc/pull/370
I don't know if there is an advantage of also setting the statement parameter?

renatocaval:
> I don't know if there is an advantage of also setting the statement parameter?

No, I don't think so. We can abandon this track.

WellingR:
I agree

PR #365, opened 2020-03-06
Added support for EventEnvelope timestamp (forwardport #346)

renatocaval:


renatocaval:
@WellingR / @patriknw, I think we lost this one from the radar. We must have it for 4.0.0

Issue #91, opened 2017-02-24
currentEventsByPersistenceId.EventEnvelope problem


skisel:
Hey,

I found a bug or confusion in currentEventsByPersistenceId query, which returns EventEnvelope and as ordering (`offset`) provides sequenceNumber which is very confusing.

JdbcReadJournal.scala:97

WellingR:
An EventEnvelope has the following signature: (as of akka 2.5)

```scala
final case class EventEnvelope(
  offset:        Offset,
  persistenceId: String,
  sequenceNr:    Long,
  event:         Any)
```

The offset is used for the eventsByTag queries, it can be ignored when using the currentEventsByPersistenceId query.

The sequenceNr is the sequence number of the message with the given persistenceId. Which can be used if you would want to execute the currentEventsByPersistenceId query again.

WellingR:
Closing this issue, if you have any more questions feel free to re-open

jroper:
@patriknw just encountered this bug myself - well, I consider it a bug, it makes no sense to me that two different events have different offsets depending on which query they were returned from. But, I can't find anything in the Akka persistence docs that defines what an offset is considered to be. My opinion is that it should be considered the offset in the event log, such that if two event envelopes have the same offset, they must be the same event, and can be correlated, cached, shared etc.

jroper:
Here's an example of when having the offset from a `currentEventsByPersistenceId` would be useful. In investigating a problem in the system, you find that the root of the problem started in a particular entity - for example, you might be doing forensics, and you find that an attacker exploited a bug that allowed them to inject an event into the event log, and by injecting that event, it allowed them to do many other things. So, you want to audit every event that occurred after that one. You're using a back office API that allows you to look at events either by persistence id, or by tag. So, you find the offset of the event that the attacker used using events by persistence id, and then you get all events by tag, starting from the offset of that.

WellingR:
This example makes sense. However the scaladoc on the definition of `Offset` is not very clear though.

renatocaval:
This is a good one to add on the upcoming 4.0.0 release.

patriknw:
For this plugin I agree that it's good to use the same global ordering offset for both eventsByPersistenceId and eventsByTag. The sequenceNumber is anyway included in the EventEnvelope if that is needed.

Also, we should make sure that the offset type used by this plugin is documented if not already.

It's a bit scary behavior change, but maybe that is fine for a 4.0.0 if we make it clear in release notes.

PR #385, opened 2020-04-17
Use ordering Offset in eventsByPersistenceId, #91

patriknw:
* to be able to correlate between eventsByPersistenceId and eventsByTag
* seqNr is anyway in the EventEnvelope
* changing signature of messages and messagesWithBatch in JournalDaoWithReadMessages
  * tuple (PersistentRepr, Long) to include the ordering number
* Scaladoc clarifications of the queries

References #91

### DIFF
--- a/core/src/main/scala/akka/persistence/jdbc/query/javadsl/JdbcReadJournal.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/query/javadsl/JdbcReadJournal.scala
@@ -25,18 +25,68 @@ class JdbcReadJournal(journal: ScalaJdbcReadJournal)
     with CurrentEventsByTagQuery
     with EventsByTagQuery {
 
+  /**
+   * Same type of query as `persistenceIds` but the event stream
+   * is completed immediately when it reaches the end of the "result set". Events that are
+   * stored after the query is completed are not included in the event stream.
+   */
   override def currentPersistenceIds(): Source[String, NotUsed] =
     journal.currentPersistenceIds().asJava
 
+  /**
+   * `persistenceIds` is used to retrieve a stream of all `persistenceId`s as strings.
+   *
+   * The stream guarantees that a `persistenceId` is only emitted once and there are no duplicates.
+   * Order is not defined. Multiple executions of the same stream (even bounded) may emit different
+   * sequence of `persistenceId`s.
+   *
+   * The stream is not completed when it reaches the end of the currently known `persistenceId`s,
+   * but it continues to push new `persistenceId`s when new events are persisted.
+   * Corresponding query that is completed when it reaches the end of the currently
+   * known `persistenceId`s is provided by `currentPersistenceIds`.
+   */
   override def persistenceIds(): Source[String, NotUsed] =
     journal.persistenceIds().asJava
 
+  /**
+   * Same type of query as `eventsByPersistenceId` but the event stream
+   * is completed immediately when it reaches the end of the "result set". Events that are
+   * stored after the query is completed are not included in the event stream.
+   */
   override def currentEventsByPersistenceId(
       persistenceId: String,
       fromSequenceNr: Long,
       toSequenceNr: Long): Source[EventEnvelope, NotUsed] =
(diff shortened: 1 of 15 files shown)


core/src/main/scala/akka/persistence/jdbc/query/javadsl/JdbcReadJournal.scala
```diff
@@ -25,34 +25,78 @@ class JdbcReadJournal(journal: ScalaJdbcReadJournal)
...
+   * Corresponding query that is completed when it reaches the end of the currently
+   * known `persistenceId`s is provided by `currentPersistenceIds`.
+   */
   override def persistenceIds(): Source[String, NotUsed] =
     journal.persistenceIds().asJava
 
+  /**
+   * Same type of query as `eventsByPersistenceId` but the event stream
+   * is completed immediately when it reaches the end of the "result set". Events that are
+   * stored after the query is completed are not included in the event stream.
```
renatocaval, line 54:
```suggestion
   * is completed immediately when it reaches the end of the "result set". New `persistenceId`s 
   * that are stored after the query is completed are not included in the stream.
```

core/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -119,12 +123,45 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
     adapter.fromJournal(repr.payload, repr.manifest).events.map(repr.withPayload)
   }
 
+  /**
+   * Same type of query as `eventsByPersistenceId` but the event stream
+   * is completed immediately when it reaches the end of the "result set". Events that are
+   * stored after the query is completed are not included in the event stream.
```
renatocaval, line 129:
```suggestion
   * is completed immediately when it reaches the end of the "result set". New `persistenceId`s 
   * that are stored after the query is completed are not included in the stream.
```

core/src/test/scala/akka/persistence/jdbc/query/QueryTestSpec.scala
```diff
@@ -178,6 +178,37 @@ class JavaDslJdbcReadJournalOperations(readJournal: javadsl.JdbcReadJournal)(
...
+      probe
+    }
+
+    private def assertEnvelope(env: EventEnvelope, persistenceId: String, sequenceNr: Long, event: Any): Unit = {
+      assert(
+        env.persistenceId == persistenceId,
+        s"expected persistenceId $persistenceId, found ${env.persistenceId}, in $env")
+      assert(env.sequenceNr == sequenceNr, s"expected sequenceNr $sequenceNr, found ${env.sequenceNr}, in $env")
+      assert(env.event == event, s"expected event $event, found ${env.event}, in $env")
+    }
```
renatocaval, line 208: +1, better visibility to what failed

patriknw:
> Maybe we should add notice in the docs about that change or were you planning to only add to the release notes?

Is there a "migration guide" section in the documentation? Otherwise I think we should just mention it in the release notes along with some of the other incompatibilities for 4.0.0.

Regarding documented behavior it was completely undefined previously unless you looked at source or ran it.

core/src/test/scala/akka/persistence/jdbc/query/QueryTestSpec.scala
patriknw: yes that is a bonus. Reason was to ignore the offset in the equals asserts.

renatocaval:
I'm fine with only mentioning on the migration guide.

There is still too open suggestion for the scaladoc. We should not mention events for those because we stream the persistent ids, not events.

core/src/main/scala/akka/persistence/jdbc/query/javadsl/JdbcReadJournal.scala
patriknw: @renatocaval ^ note currentEventsByPersistenceId vs currentPersistenceIds

renatocaval: oh, sorry. 🤦

PR #390, opened 2020-04-23
Update to Akka 2.6.4 and ScalaTest 3.1.1

patriknw:
* removing support for Akka 2.5.x because we can't run persistence-tck
  tests with both Akka 2.5 and 2.6 due to different ScalaTest versions
* for the 4.0.0 release we don't want to support Akka 2.5.x anyway,
  users that need to stay on Akka 2.5 can use the old akka-persistence-jdbc 3.5.x
  version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ jobs:
     - stage: test
       env:
         - PRE_CMD="./scripts/launch-all.sh"
-        - CMD="++2.12.10 test"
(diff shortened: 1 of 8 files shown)


ennru:
I'll remove the Travis cron job (which did the Akka 2.6 build).

patriknw:
I had messed up the patience in JournalSequenceActorTest, should be fixed now

Issue #400, opened 2020-05-18
Slow quering the highest sequence number in the journal table on Oracle DB


samfrown:
### Versions used

akka-persistence-jdbc: 3.5.0
slick: 3.3.2

Akka version: 2.5.23

### Expected Behavior

Predictable time for an actor persistent state loading on restart from Oracle DB in production.

### Actual Behavior

We are using akka-persistence to store state in Oracle Database.

Our average load is about 80M events per day.

We don't use hard deletion of events as it greatly increase database load. Instead we use database side job which clean up data by dropping outdated table partition daily.
The "journal" table is partitioned by "creation_time" column and we could safely drop a previous day partition after the first current day snapshot is saved.

So, lately we faced a problem of very slow journal loading when the actor was restarted at the end of a day.

During investigation we found that most time during loading actor spent in the following query:

```
select x2.x3 from (select "sequence_number" as x3 from "journal"
where "persistence_id" = :1 order by "sequence_number" desc) x2
where rownum <= 1
```

Caused by the nested query with the "order by ... desc" statement this query tends to have a non-optimal execution plan for the table with significant amount of records.

In our production environment we measured up to 10 minutes execution time for it.

To fix this issue we changed a query code in [JournalQueries.scala](https://github.com/akka/akka-persistence-jdbc/compare/3.5.x...samfrown:optimize_query_of_highest_seq_number) which produce an sql like this:
```
select max("sequence_number") from "journal" where "persistence_id" = :1
```

This query has simplier execution plan and has more constant evaluation time regardless of the size on the table.

### Relevant logs

n/a

### Reproducible Test Case

Oracle Database 12c
Number of events: 80 000 000
Size of the event: 100 bytes

WellingR:
I had a quick and these changes make sense. Could you create pull request for this?

octonato:
Done in #401

PR #401, opened 2020-05-19
Optimize query of highest seq number

samfrown:
References #400 

Change highest sequence number queries to simplify generated sql

### DIFF
--- a/core/src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalDao.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/journal/dao/ByteArrayJournalDao.scala
@@ -134,11 +134,11 @@ trait BaseByteArrayJournalDao extends JournalDaoWithUpdates with BaseJournalDaoW
   }
 
   private def highestMarkedSequenceNr(persistenceId: String) =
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

renatocaval:
@samfrown, thanks for this contribution. 

We are about to release v4.0.0 and I would like to suggest that you re-target this fix to the `master` branch instead of `3.5.x`. We are not planning to release from `3.5.x` branch anymore, unless serious bug is found. 

The release process and groupId changed since it moved under the Akka org and it make it much easier for us to not keep releasing old branches.

We can include your fix in `v4.0.1`. 
The reason it's preferable to not include it in `4.0.0`, is that we have already cut a RC and since this is not a critical bug fix, we prefer that `4.0.0` to be the same as `4.0.0-RC2`. I hope you understand.

We can cut v4.0.1 short after.

samfrown:
@renatocaval Finally I have been able to rebase the branch as you asked

renatocaval:
Thanks @samfrown

PR #441, opened 2020-10-08
poolName defined for different connection pools

Zuchos:
When DB configuration is reused it's wise to set up different connection pool name for monitoring/debugging purposes. I would said that having 3 connection pools with same name is bad practice, so it should not be in reference conf.

### DIFF
--- a/core/src/test/resources/postgres-application.conf
+++ b/core/src/test/resources/postgres-application.conf
@@ -32,16 +32,19 @@ akka {
 
 jdbc-journal {
   slick = ${slick}
+  slick.db.poolName = "write-journal"
 }
 
 # the akka-persistence-snapshot-store in use
 jdbc-snapshot-store {
   slick = ${slick}
+  slick.db.poolName = "snapshot-store"
 }
 
 # the akka-persistence-query provider in use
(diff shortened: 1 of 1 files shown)


lightbend-cla-validator:
Hi @Zuchos,

octonato (commented):
I think the goal here is to have one single connection pool for all there underlying components.

Moreover, those are the files uses in tests, not for user consumption. Have you run into an issue in production in which using names would have helped you? 

Maybe instead of changing the config files we use in tests, we should document how to setup the names in prod code. WDYT?

WellingR:
These files are not only for the tests, but we also link to this file in the documentation:
https://doc.akka.io/docs/akka-persistence-jdbc/current/#sharing-the-database-connection-pool-between-the-journals

I guess it makes sense to configure these settings for all databases, so that our documentation includes the pool names in the examples

octonato:
In that case, we should have for all DBs and also for the shared db settings.

patriknw:
@Zuchos Is this something you would like to complete or shall we close? Then it should include that setting for the other database configuration also.

ennru:
Closing for inactivity.

Issue #451, opened 2020-11-06
JdbcReadJournal doesn't use the configuration passed in parameters to resolve the write plugin


rdubois:
Please read https://github.com/akka/akka/issues/28934 and the underlying rejected PR https://github.com/akka/akka/pull/28935.

octonato:
Solved in #453

PR #452, opened 2020-11-06
fix(#451): provide the plugin configuration when looking for the eventAdapters

rdubois:
References #451

The write-plugin is resolved using the default actor system configuration and then doesn't support an eventual configuration provided at runtime.

The PR resolves this issue by providing the runtime plugin configuration while fetching the event adapters. If the plugin configuration is empty, or if the plugin reference is not found, it will be resolved using the config of the actor system, for retro-compatibility.

### DIFF
--- a/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
+++ b/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
@@ -73,7 +73,10 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
   val readJournalConfig = new ReadJournalConfig(config)
 
   private val writePluginId = config.getString("write-plugin")
-  private val eventAdapters = Persistence(system).adaptersFor(writePluginId)
+
+  // If 'config' is empty, or if the plugin reference is not found, then the write plugin will be resolved from the
+  // ActorSystem configuration. Otherwise, it will be resolved from the provided 'config'.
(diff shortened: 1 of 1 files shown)


patriknw (approved):
LGTM. Can you use version 4.0.x instead of 3.5.x? In that case this should be targeted to `master` instead.

rdubois:
Hi @patriknw, I will do another PR which will target master yes. Could we keep both, so that this change will be available for akka 2.5.X ?

rdubois:
Here is the PR which targets master: https://github.com/akka/akka-persistence-jdbc/pull/453

octonato:
We should close this one, since we won't be cutting 3.5.x releases anymore. This is due to the change of repo org. We (Lightbend) don't have the rights to release on the old groupId. 

It's possible with some cooperation from previous maintainer, but we should only do it in case of important bug fixes.

@rdubois, are you using a 3.5.x release? Would it be possible for you to upgrade to 4.0.0? Note that the artifact groupId has changed.

octonato:
> Could we keep both, so that this change will be available for akka 2.5.x ?

@rdubois sorry, I missed your comment above.

Akka 2.5.x reached EOL (see our most recent [blog post](https://akka.io/blog/news/2020/11/16/one-year-with-akka-2-6)) and we are not planning any updates to support it. Unless it's about a important security bug fixes. 

True that this is not about Akka 2.5.x, but still related and on top of that, we are not cutting 3.5.x release anymore (as explained above).

rdubois:
@octonato thanks for the prompt answer. It makes sense. Let's close that one then.

PR #453, opened 2020-11-17
fix(#451): provide the plugin configuration when looking for the eventAdapters

rdubois:
References #451

The write-plugin is resolved using the default actor system configuration and then doesn't support an eventual configuration provided at runtime.

The PR resolves this issue by providing the runtime plugin configuration while fetching the event adapters. If the plugin configuration is empty, or if the plugin reference is not found, it will be resolved using the config of the actor system, for retro-compatibility.


octonato (approved):

octonato:
Thanks @rdubois

Issue #69, opened 2016-09-02
Option to auto create tables


jroper:
It would be very convenient for users in development and testing (and perhaps even in production) if akka-persistence-jdbc offered the option to auto create tables if they don't exist.  akka-persistence-cassandra offers this for example:

https://github.com/akka/akka-persistence-cassandra/blob/master/src/main/resources/reference.conf#L52

Slick can automagically generate and execute database specific DDL, so this shouldn't be too hard.  I'd imagine it would be sufficient to simply check if each table exists first, and if it doesn't execute the create.  To deal with cluster scenarios with two nodes creating the tables at once, if the creation fails you could then check if the table has been created after the failure, and if it has then assume the failure was because another node created it at the same time and so ignore it, otherwise rethrow.

dnvriend:
Hi! This seems like a good addition to the plugin. If you feel up to it, you could make the change and do a PR.

jroper:
I'm currently implementing this myself in Lagom, it probably won't be hard to move what I've got to here once I'm finished.

There are a few things to consider though, firstly you have to detect whether the tables are already created, this turns out to be a little harder than it sounds (if you get a list of tables, they may have a schema in them which could be the current schema, so you need to have some dialect specific way to work out the current schema to do the matching).  You also need to deal with multiple nodes creating the tables at the same time, one possibility is to use a cluster singleton to create them, another is if creation fails, check if the tables exist after the failure, and if they do, ignore the failure.

Fristi:
I also wanted this in a project. I am using single node with h2 (in mem atm) for testing. It seems the current schema will hang upon creation with the combo of h2 and slick

There are two problems
- Snapshot schema uses `O.AutoInc` on two columns and the combination slick + h2 seems to dislike that
- Journal uses `O.AutoInc` in conjunction with primary key which also causes to hang. Because `O.AutoInc` is a primary key automatically (https://github.com/slick/slick/issues/763)

I copied the schema sources and removed these lines and it solved it for me :-).

As @jroper states, in a production environment with a real DB this would cause issues, but just posting this for a head up.

dnvriend:
@Fristi Thanks for the head up!

stevemyers:
I used Slick to create my H2 akka-persistence-jdbc schema, I know it's not a full solution but works good for me as an interim solution.  See gist [here](https://gist.github.com/stevemyers/de1fe4db29555eaea83fa1df3d13a5af)

[...]

PR #465, opened 2020-12-10
Introduces SchemaUtils to programatically drop and create schemas

octonato:
References #69


core/src/main/scala/akka/persistence/jdbc/schema/scaladsl/SchemaUtils.scala
```diff
@@ -0,0 +1,126 @@
...
+      val matchingProfile = "slick.jdbc.OracleProfile$"
+    }
+    case object SqlServer extends SchemaType {
+      val schema = "schema/sqlserver/sqlserver-schema.sql"
+      val separator = ";"
+      val matchingProfile = "slick.jdbc.SQLServerProfile$"
+    }
+  }
+
+  def dropAndCreate(implicit actorSystem: ActorSystem): Future[Done] = {
```
ennru, line 61: Make it compatible with the new actors API out-of-the-box:
```suggestion
  def dropAndCreate(implicit actorSystem: ClassicActorSystemProvider): Future[Done] = {
```

ennru (approved):
LGTM.

octonato:
@ennru, a second review round is welcome. 

I did some more clean-ups, added Java API, API docs and updated the documentation.

docs/src/main/paradox/index.md
```diff
@@ -59,11 +59,32 @@ Configure `slick`:
...
+import akka.persistence.jdbc.schema.scaladsl.SchemaUtils
+implicit val system: ActorSystem = ActorSystem("example")
+val done: Future[Done] = SchemaUtils.dropAndCreate()
+```
+
+Java
+:  ```java
+import akka.persistence.jdbc.schema.javadsl.SchemaUtils;
+ActorSystem actorSystem = ActorSystem.create("example")
+CompletionStage<Done> done = SchemaUtils.dropAndCreate(actorSystem)
```
octonato, line 83: we need to start using imports from code snippets, but I'm explicitly leaving that for another occasion so we can move faster with this PR.

core/src/main/scala/akka/persistence/jdbc/schema/scaladsl/SchemaUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+   * For information about the different schemas and supported database can be found at
+   * https://doc.akka.io/docs/akka-persistence-jdbc/current/index.html#database-schema
+   *
+   * This method will automatically detect the configured database using the settings from `jdbc-journal` config.
+   * If configured with `use-shared-db`, it will use the `akka-persistence-jdbc.shared-databases` definition instead.
+   * See https://doc.akka.io/docs/akka-persistence-jdbc/current/index.html#sharing-the-database-connection-pool-between-the-journals for details.
+   *
+   * @param systemProvider a ClassicActorSystemProvider
+   */
+  def dropAndCreate(implicit systemProvider: ClassicActorSystemProvider): Future[Done] = {
```
johanandren, line 44: should it be `dropAndCreate()(implicits)` given that it is not a pure method, so that usage becomes `dropAndCreate()`?

core/src/main/scala/akka/persistence/jdbc/schema/scaladsl/SchemaUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+   *
+   * To use the snapshot configuration, pass `jdbc-snapshot-store` to `configKey`.
+   *
+   * @param schema the DDL script file. Must be accessible from the classpath.
+   * @param separator used to separate the different DDL statement. Defaults to ';' for most databases.
+   * @param configKey the configuration key to use, default to `jdbc-journal`
+   * @param systemProvider a ClassicActorSystemProvider
+   * @return
+   */
+  def applySchema(schema: String, separator: String = ";", configKey: String = "jdbc-journal")(
```
johanandren, line 79: Default params are problematic for bincomp, not sure what we promise around that for akka-persistence-jdbc?

core/src/main/scala/akka/persistence/jdbc/schema/scaladsl/SchemaUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+        ddl <- for {
+          trimmedLine <- schema.split(separator).map(_.trim)
+          if trimmedLine.nonEmpty
+        } yield trimmedLine
+      } withStatement(db) { stmt =>
+        logger.debug(s"applying DDL: $ddl")
+        try stmt.executeUpdate(ddl)
+        catch {
+          case t: SQLSyntaxErrorException if shouldSuppressKnownError(t.getMessage) =>
+          case t: java.sql.SQLException if shouldSuppressKnownError(t.getMessage)   =>
```
johanandren, line 108: Not worth logging at debug because we are so sure they are never interesting?

core/src/test/scala/akka/persistence/jdbc/journal/JdbcJournalPerfSpec.scala
```diff
@@ -46,7 +51,7 @@ abstract class JdbcJournalPerfSpec(config: Config, schemaType: SchemaType)
   lazy val db = SlickExtension(system).database(cfg).database
 
   override def beforeAll(): Unit = {
-    dropCreate(schemaType)
+    Await.result(SchemaUtils.dropAndCreate(system), 2.seconds)
```
johanandren, line 54: .futureValue is available because `ScalaFutures` but perhaps we don't want the patience config for that here?

johanandren (approved):
LGTM, with some nitpicks

core/src/main/scala/akka/persistence/jdbc/schema/scaladsl/SchemaUtils.scala
ennru: We do, this should be done with an overload instead.

core/src/main/scala/akka/persistence/jdbc/schema/scaladsl/SchemaUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  private def applySchema(
+      db: Database,
+      schema: String,
+      separator: String,
+      blockingEC: ExecutionContext): Future[Done] = {
+    Future {
+      for {
+        schema <- Option(fromClasspathAsString(schema))
```
ennru, line 98: Why is this wrapped in an option? How does it fail if the schema/filename can't be read as a resource?
Should there be an API to provide the DDL directly?

ennru (approved):
Looks still good!
A few suggestions.

core/src/main/resources/schema/oracle/oracle-schema.sql
```diff
@@ -1,3 +1,24 @@
...
+
+DROP TRIGGER "ordering_seq_trigger"
+/
+
+DROP PROCEDURE "reset_sequence"
+/
+
+DROP SEQUENCE "ordering_seq"
+/
+
```
ignasi35, line 21: I know this aligns the script to those for MySQL, Postgre, etc... where there is a DROP first and CREATE later, but this PR is also moving the `.sql` scripts to `main/` (from `test/`).

I'd rather we had (ideally idempotent) `.sql` scripts in `main/` that only created and then, somewhere else (e.g. `test/`) the scripts to drop.

ignasi35: As a side effect of that split, we would be improving some cases (for example postgre-schema.sql) where there is:

```sql
DROP TABLE IF EXISTS public.journal;

CREATE TABLE IF NOT EXISTS public.journal (
...
);

DROP TABLE IF EXISTS public.snapshot;

CREATE TABLE IF NOT EXISTS public.snapshot (
...
```

Which looks a bit weird ("create table if not exists" immediately after "drop table"?).

core/src/main/scala/akka/persistence/jdbc/schema/javadsl/SchemaUtils.scala
```diff
@@ -0,0 +1,81 @@
...
+import akka.persistence.jdbc.db.SlickExtension
+import slick.jdbc.JdbcBackend.Database
+
+object SchemaUtils {
+
+  /**
+   * JAVA API:
+   *
+   * For testing purposes the journal and snapshot tables can be created programmatically.
+   * For production it's recommended to create the table with DDL statements before the system is started.
```
ignasi35, line 27: I think it'd be great to have a dedicated module for these tools. Getting them in production code, out of the box looks weird.

core/src/main/scala/akka/persistence/jdbc/schema/javadsl/SchemaUtils.scala
```diff
@@ -0,0 +1,81 @@
...
+import akka.persistence.jdbc.schema.scaladsl
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.Future
+
+import akka.dispatch.Dispatchers
+import akka.persistence.jdbc.db.SlickDatabase
+import akka.persistence.jdbc.db.SlickExtension
+import slick.jdbc.JdbcBackend.Database
+
+object SchemaUtils {
```
ignasi35, line 21:
```suggestion
object TestingUtils {
```

core/src/main/scala/akka/persistence/jdbc/schema/javadsl/SchemaUtils.scala
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright (C) 2014 - 2019 Dennis Vriend <https://github.com/dnvriend>
+ * Copyright (C) 2019 - 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.persistence.jdbc.schema.javadsl
```
ignasi35, line 6:
```suggestion
package akka.persistence.jdbc.testkit.javadsl
```

core/src/main/scala/akka/persistence/jdbc/schema/scaladsl/SchemaUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+import akka.persistence.jdbc.db.SlickExtension
+import org.slf4j.LoggerFactory
+import slick.jdbc.H2Profile
+import slick.jdbc.JdbcBackend.Database
+import slick.jdbc.MySQLProfile
+import slick.jdbc.OracleProfile
+import slick.jdbc.PostgresProfile
+import slick.jdbc.SQLServerProfile
+
+object SchemaUtils {
```
ignasi35, line 28: Same rename suggestions.

core/src/test/scala/akka/persistence/jdbc/SingleActorSystemPerTestSpec.scala
```diff
@@ -70,6 +76,8 @@ abstract class SingleActorSystemPerTestSpec(val config: Config)
 
   def withActorSystem(f: ActorSystem => Unit): Unit = {
     implicit val system: ActorSystem = ActorSystem("test", config)
+    // start new schema on each system creation and for each individual test
+    Await.ready(SchemaUtils.dropAndCreate(system), 2.seconds)
```
ignasi35, line 80:
```suggestion
    SchemaUtils.dropAndCreate(system).futureValue
```

use `futureValue` from `org.scalatest.concurrent.ScalaFutures`?

ignasi35: Ah, just noticed @johanandren's comment about patience config.

ignasi35 (changes requested):
I'd rather make it clearer the new `SchemaUtils` is not meant for production. Either renaming the object and/or the package, or even moving the code and SQL scripts into a new module users can import with scope `Test`.

ignasi35:
> I'd rather make it clearer the new `SchemaUtils` is not meant for production. Either renaming the object and/or the package, or even moving the code and SQL scripts into a new module users can import with scope `Test`.

This suspiciously reminded me of https://github.com/lagom/lagom/blob/master/persistence-cassandra/javadsl/src/main/scala/com/lightbend/lagom/javadsl/persistence/cassandra/testkit/TestUtil.scala

core/src/main/scala/akka/persistence/jdbc/schema/javadsl/SchemaUtils.scala
patriknw: maybe a new module (jar artifact) is overkill, but just by placing them in a `testkit` package would clarify expectations

core/src/main/scala/akka/persistence/jdbc/schema/scaladsl/SchemaUtils.scala
patriknw: I'd say that bincompat is a lost promise for this plugin until we rewrite it, but good to not introduce more warts

core/src/main/scala/akka/persistence/jdbc/schema/scaladsl/SchemaUtils.scala
patriknw: why two so different names, `dropAndCreate` vs `applySchema` ?

core/src/test/scala/akka/persistence/jdbc/serialization/StoreOnlySerializableMessagesTest.scala
```diff
@@ -54,7 +56,7 @@ abstract class StoreOnlySerializableMessagesTest(config: String, schemaType: Sch
   }
 
   override def beforeAll(): Unit = {
-    dropCreate(schemaType)
+    Await.result(SchemaUtils.dropAndCreate(system), 2.seconds)
```
patriknw, line 59: why so tight timeout?

patriknw (commented):
looking good

core/src/main/scala/akka/persistence/jdbc/schema/scaladsl/SchemaUtils.scala
octonato: > why two so different names, dropAndCreate vs applySchema ?

Because the default schema is a drop and create. 

The `applySchema` can be any DDL script.

Maybe better if we call if `applyDefaultSqlScript` and `applySqlScript`. 

The passed script can be anything you want. It can be an alter table, a drop, inserts.

core/src/main/scala/akka/persistence/jdbc/schema/scaladsl/SchemaUtils.scala
octonato: Yes, will do it

core/src/test/scala/akka/persistence/jdbc/journal/JdbcJournalPerfSpec.scala
octonato: that's how we did everywhere in akka-projections.

core/src/main/scala/akka/persistence/jdbc/schema/scaladsl/SchemaUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  private val logger = LoggerFactory.getLogger("akka.persistence.jdbc.schema.SchemaUtils")
```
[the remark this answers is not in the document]
octonato: I was in doubt about this one. The javadsl relies on the scaladsl implementation. So, if you are using Java, you will have to configure the logging using `.scaladsl.`. 

Confusing, but also confusing the way I did. Maybe better to have a `.internal.SchemaUtils` to which we pass the logger from the Scala/Java one. So nobody gets surprised.

core/src/test/scala/akka/persistence/jdbc/serialization/StoreOnlySerializableMessagesTest.scala
octonato: I'll can increase, but I find it high already.

core/src/main/scala/akka/persistence/jdbc/schema/javadsl/SchemaUtils.scala
octonato: I will move it to `testkit` although I don't think it's necessary. 

I don't think it's such a bad thing to not have it as is. We do the same in projection. The `createIfNotExists` methods are on the respective Projection companions. 

We have the scaladoc saying we don't recommend it. The method name is super explicit about it, it's `dropAndCreate`.

core/src/main/scala/akka/persistence/jdbc/schema/javadsl/SchemaUtils.scala
octonato: Not sure if we should to it. Isn't `testkit.SchemaUtils` good enough?

core/src/main/resources/schema/oracle/oracle-schema.sql
octonato: Ok to remove the `IF NOT EXISTS` from create table. 

Since the goal is to use it as a helper for building test, I do think it should drop and create by default. Users building tests do need to drop. If they don't want to drop the schema, they can use `applySchema` instead and pass a another file.

core/src/main/scala/akka/persistence/jdbc/schema/javadsl/SchemaUtils.scala
patriknw: `testkit` package is enough in my opinion

core/src/main/scala/akka/persistence/jdbc/schema/scaladsl/SchemaUtils.scala
octonato: I copied this code from tests. I guess the reason is that we are flatMapping with another option, so we need to align. It could surely be different. 

> Should there be an API to provide the DDL directly?

I didn't want to increase the scope and the options. I found it was a good compromise to require the schema to be on the classpath, like in `src/test/resources`.

Can be added later I think.

core/src/test/scala/akka/persistence/jdbc/serialization/StoreOnlySerializableMessagesTest.scala
patriknw: Such a waste to spend time on flaky tests in slow travis or similar env. It will not test better by a short timeout here.

core/src/main/resources/schema/oracle/oracle-schema.sql
octonato: I was thinking about what I just said here. 

Maybe we should indeed split the scripts and have: `drop`, `create`, `dropAndCreate` and `applyScript`. The later being the generic one.

core/src/test/scala/akka/persistence/jdbc/journal/JdbcJournalPerfSpec.scala
octonato: I went dig into akka-projections to remember why we did so. Here the comment we have there

```scala
// don't use futureValue (patience) here because it can take a while to start the test container
```

If I remember correctly, creating the DB can take much longer than other async in tests and we didn't want to increase the patience config because of that. So, for the specific case of creating DB we start to use the general Scala API.

core/src/main/scala/akka/persistence/jdbc/testkit/internal/SchemaUtilsImpl.scala
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  def dropAndCreateSchema(logger: Logger)(implicit actorSystem: ClassicActorSystemProvider): Future[Done] =
+    dropSchema(logger).flatMap(_ => createSchema(logger))(actorSystem.classicSystem.dispatcher)
+
+  def applyScript(script: String, separator: String, configKey: String, logger: Logger)(
+      implicit actorSystem: ClassicActorSystemProvider): Future[Done] =
+    applyScript(script, separator, logger, loadSlickDatabase(configKey))
+
+  private def applyScript(schema: String, separator: String, logger: Logger, slickDb: SlickDatabase)(
```
octonato, line 60: @ennru, at the end I did what you suggested. One can pass the whole script in a String and it gets applied. 

The thing is, if you want to load something from the classpath, better to read the file into a string and pass to this method. 

Note that the `applyScript` is the power-user variant, you are on your own to pass all the bits.

octonato (commented):
I applied almost all suggestion, plus some other clean-ups:

* use `()` to indicate side-effect
* log in debug in case of errors
* package `testkit`
* `internal.SchemaUtilsImpl` to not create logging confusion
* split scripts into `drop` and `create`
* `applyScript` not dependent on file classpath

octonato:
One more thing, not all schema file has a proper `IF NOT EXISTS`. I need to search more about it, but it seems that this can different a lot.

docs/src/main/paradox/index.md
```diff
@@ -59,11 +59,32 @@ Configure `slick`:
...
+import akka.persistence.jdbc.testkit.scaladsl.SchemaUtils
+implicit val system: ActorSystem = ActorSystem("example")
+val done: Future[Done] = SchemaUtils.dropAndCreateSchema()
+```
+
+Java
+:  ```java
+import akka.persistence.jdbc.testkit.javadsl.SchemaUtils;
+ActorSystem actorSystem = ActorSystem.create("example")
+CompletionStage<Done> done = SchemaUtils.dropAndCreateSchema(actorSystem)
```
ennru, line 83:
```suggestion
ActorSystem actorSystem = ActorSystem.create("example");
CompletionStage<Done> done = SchemaUtils.dropAndCreateSchema(actorSystem);
```

ennru (approved):
This grew a bit over these days, good job @octonato!

core/src/test/resources/logback-test.xml
```diff
@@ -37,6 +37,7 @@
 
     <!--<logger name="slick.util.AsyncExecutor" level="DEBUG" />-->
 
+    <logger name="akka.persistence.jdbc.schema.scaladsl.SchemaUtils" level="DEBUG"/>
```
octonato, line 40:
```suggestion
    <logger name="akka.persistence.jdbc.testkit.scaladsl.SchemaUtils" level="DEBUG"/>
```

core/src/main/scala/akka/persistence/jdbc/testkit/javadsl/SchemaUtils.scala
```diff
@@ -0,0 +1,122 @@
...
+   *
+   * This utility method is intended to be used for testing only.
+   * For production, it's recommended to run any DDL statements before the system is started.
+   *
+   * It will use the database settings found under `jdbc-journal`, or `akka-persistence-jdbc.shared-databases` if configured so.
+   * See https://doc.akka.io/docs/akka-persistence-jdbc/current/index.html#sharing-the-database-connection-pool-between-the-journals for details.
+   *
+   * @param script the DDL script. The passed script can contain more then one SQL statements separated by a ; (semi-colon).
+   * @param actorSystem a ClassicActorSystemProvider
+   * @return
```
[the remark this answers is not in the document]
octonato: why?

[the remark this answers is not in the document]
ennru: Having docs without content doesn't help the user. So either write something useful or remove them.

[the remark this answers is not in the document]
octonato: oh, you mean the `@return`? I thought you were talking about the whole docs.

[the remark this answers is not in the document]
ennru: The `@return` without docs should go, and `@param actorSystem` should either go or say `an actor system, classic och new API`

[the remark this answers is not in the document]
octonato: done

octonato:
@ennru. I did one more change here after you last approval. 

See my comment in https://github.com/akka/akka-platform-guide/pull/422#issuecomment-745551261

In a nutshell, I removed the `dropAndCreateSchema`. Instead, we should use `dropIfExists` and `createIfNotExists`. This reflects better what it happens behind. 

I will adapt projection to have the same.

octonato:
> I will adapt projection to have the same.

Done. See [`e500e7e` (#490)](https://github.com/akka/akka-projection/pull/490/commits/e500e7eebdf7f62f3166c99ac5a842beb1452677)

ignasi35 (approved):

octonato:
@ennru, this should pass on first run (famous last words).

`DropCreate` is re-introduced and now relies on `internal.SchemaUtilsImpl` to do its job. `SchemaUtilsImpl` exposes a little bit more (internal API) in order to allow it. 

Previous changes I did was too aggressive and was causing Oracle tests to fail. That's why `DropCreate` and `Cleaners` are not back.

Main changes are in `DropCreate` and `SchemaUtilsImpl`, the rest is are basically the same.

Issue #473, opened 2020-12-17
Tweak build to fail on warnings


octonato:
Like any other project in the Akka org. 

There are a few warnings to remove as of today, but feasible.

PR #478, opened 2021-01-08
Make Scalac complain more

ennru:
Not all the way to fatal warnings, but most warnings are fixed.
I didn't want to bring in the Silencer in this round.

- enable more compiler warnings
- fix most compiler warnings

References #473

### DIFF
--- a/core/src/main/scala/akka/persistence/jdbc/journal/dao/legacy/ByteArrayJournalDao.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/journal/dao/legacy/ByteArrayJournalDao.scala
@@ -5,12 +5,7 @@
 
 package akka.persistence.jdbc.journal.dao.legacy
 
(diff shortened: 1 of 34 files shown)


core/src/main/scala/akka/persistence/jdbc/journal/dao/legacy/ByteArrayJournalDao.scala
```diff
@@ -118,7 +110,7 @@ trait BaseByteArrayJournalDao
     val write = PersistentRepr(payload, sequenceNr, persistenceId)
     val serializedRow = serializer.serialize(write) match {
       case Success(t) => t
-      case Failure(ex) =>
+      case Failure(_) =>
```
chbatey, line 113: maybe put as the cause instead?

chbatey (approved):
LGTM with one commend

octonato (approved):

PR #487, opened 2021-01-18
Use section akka.persistence.jdbc (instead of akka-persistence-jdbc)

ennru:
It is an anomaly of this plugin to read settings from the root config section `akka-persistence-jdbc` where it should have been a subsection as `akka.persistence.jdbc`.

As version 5.0 breaks the database schema which will make it unlikely for users to migrate from earlier versions, changing the config structure could be an option.


patriknw (commented):
looks simple enough, if we don't care about breaking user's config (and I'm fine with that for 5.0.0)

PerWiklander:
This looks dead? At least 5.0.0 has come and gone.

ennru:
> This looks dead? At least 5.0.0 has come and gone.

I agree, I forgot about this one. Maybe not that much of a problem either.

Issue #371, opened 2020-03-26
Improve how config default values are handled


patriknw:
Default values should be in reference.conf and not defined in code.
I found that it is not even reading some config values from reference.conf. For example the batchSize in BaseByteArrayJournalDaoConfig.

ennru:
Fixed with #488

PR #488, opened 2021-01-19
Read config exclusively from config files

ennru:
Fail if code uses a default value specified in the source code.
All configuration has to be present via config.

References #371

### DIFF
--- a/core/src/main/scala/akka/persistence/jdbc/config/AkkaPersistenceConfig.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/config/AkkaPersistenceConfig.scala
@@ -15,132 +15,132 @@ object ConfigKeys {
 }
 
 class SlickConfiguration(config: Config) {
-  val jndiName: Option[String] = config.as[String]("jndiName").trim
-  val jndiDbName: Option[String] = config.as[String]("jndiDbName")
+  val jndiName: Option[String] = config.asOptionalNonEmptyString("jndiName")
+  val jndiDbName: Option[String] = config.asStringOption("jndiDbName")
   override def toString: String = s"SlickConfiguration($jndiName,$jndiDbName)"
 }
 
 class LegacyJournalTableColumnNames(config: Config) {
-  private val cfg = config.asConfig("tables.legacy_journal.columnNames")
-  val ordering: String = cfg.as[String]("ordering", "ordering")
-  val deleted: String = cfg.as[String]("deleted", "deleted")
-  val persistenceId: String = cfg.as[String]("persistenceId", "persistence_id")
-  val sequenceNumber: String = cfg.as[String]("sequenceNumber", "sequence_number")
-  val created: String = cfg.as[String]("created", "created")
-  val tags: String = cfg.as[String]("tags", "tags")
-  val message: String = cfg.as[String]("message", "message")
+  private val cfg = config.getConfig("tables.legacy_journal.columnNames")
+  val ordering: String = cfg.getString("ordering")
+  val deleted: String = cfg.getString("deleted")
+  val persistenceId: String = cfg.getString("persistenceId")
+  val sequenceNumber: String = cfg.getString("sequenceNumber")
+  val created: String = cfg.getString("created")
+  val tags: String = cfg.getString("tags")
+  val message: String = cfg.getString("message")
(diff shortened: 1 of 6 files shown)


core/src/main/scala/akka/persistence/jdbc/util/ConfigOps.scala
```diff
@@ -55,11 +56,15 @@ object ConfigOps {
...
 
-    def ?:[A](key: String, default: A) = as(key, default)
+    def ?:[A](key: String, default: A) = as(key, noDefault(key))
 
     def withkey[A](key: String)(f: Config => A): A = f(config.getConfig(key))
   }
 
+  private def noDefault(key: String) = {
+    throw new ConfigurationException(s"The [$key] setting relies on a default provided in code.")
+  }
```
patriknw, line 66: I think it would be better to use the same approach as we use in other Akka libraries, i.e. no such Try stuff if it's reading a config property that is expected to be defined in reference.conf.
But this might be a step in the right direction.

chbatey: 👍 double 

Great to get rid of the double defaults

core/src/main/scala/akka/persistence/jdbc/util/ConfigOps.scala
```diff
@@ -55,11 +56,15 @@ object ConfigOps {
...
     def ?[A](key: String): Try[A] = as(key)
 
-    def ?:[A](key: String, default: A) = as(key, default)
+    def ?:[A](key: String, default: A) = as(key, noDefault(key))
 
     def withkey[A](key: String)(f: Config => A): A = f(config.getConfig(key))
   }
 
+  private def noDefault(key: String) = {
+    throw new ConfigurationException(s"The [$key] setting relies on a default provided in code.")
```
octonato, line 65: maybe here we should say: "[$key] not found" and then remove the `default` arg?

core/src/main/scala/akka/persistence/jdbc/util/ConfigOps.scala
ennru: Yes, the possibility to put in defaults should be removed altogether.
So far this PR is mainly to show what was relying on these defaults.

core/src/main/scala/akka/persistence/jdbc/config/AkkaPersistenceConfig.scala
```diff
@@ -15,132 +15,132 @@ object ConfigKeys {
 }
 
 class SlickConfiguration(config: Config) {
-  val jndiName: Option[String] = config.as[String]("jndiName").trim
-  val jndiDbName: Option[String] = config.as[String]("jndiDbName")
+  val jndiName: Option[String] = config.asOptionalNonEmptyString("jndiName")
+  val jndiDbName: Option[String] = config.asStringOption("jndiDbName")
```
octonato, line 19: I think the second one `jndiDbName` was missing the `.trim`. I don't see a reason to a have a empty string as a valid value for the `jndiDbName`. That said, we can have one single method that always return `None` for empty strings. 

And since it's internal API, we can keep its name short, eg: `asStringOption`

core/src/test/scala/akka/persistence/jdbc/configuration/ConfigOpsTest.scala
```diff
@@ -11,63 +11,21 @@ import ConfigOps._
...
-    cfg.as[String]("person.firstName", "john") shouldBe "john"
-    cfg.as[String]("person.lastName", "doe") shouldBe "doe"
-    cfg.as[Int]("person.age", 35) shouldBe 25
-    cfg.as[Boolean]("person.hasCar", false) shouldBe false
-    cfg.as[Boolean]("person.hasGirlfriend", false) shouldBe true
+    cfg.asStringOption("person.firstName").get shouldBe "foo"
+    cfg.asStringOption("person.lastName").get shouldBe "bar"
+    cfg.asOptionalNonEmptyString("person.pet") shouldBe None
+    cfg.asOptionalNonEmptyString("person.car") shouldBe None
+    cfg.asStringOption("person.bike") shouldBe None
```
octonato, line 28: We are missing a test that confirms that `asStringOption` returns `Some("")` for `""`, but as argued previously I think it should return `None`.

octonato (approved):
LGTM with a minor suggestion

PR #496, opened 2021-01-22
Don't update Akka via scalasteward

ignasi35:
References https://github.com/akka/akka-persistence-jdbc/pull/492#issuecomment-765323017


octonato (approved):
Thanks @ignasi35. 

I was looking for the steward syntax for this.

PR #499, opened 2021-01-22
Run checks with GitHub Actions

octonato:
This is a first step PR.

It adds the preliminary checks using GitHub Actions (code format, header, mima, docs). 

Unfortunately, GH Actions don't allow the restart of a single job. We have lots of tests, sometime heavy tests, depending on docker. That means that if a test fail, we have to re-run the whole workflow (all jobs) again.

Therefore, I want to try to use workflow dependency, but first we need to have the main workflow in the master branch. As explained here: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run

Travis is still enabled and run all tests.

### DIFF
--- a/.github/workflows/check-and-build.yml
+++ b/.github/workflows/check-and-build.yml
@@ -1,0 +1,90 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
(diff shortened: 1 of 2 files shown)


seglo (approved):
LGTM

PR #501, opened 2021-02-05
Better SchemaUtils and Data Migration tool

Tochemey:
The goal of the PR is to provide:

- a flyway-based schemas creation tool
- a data migration tool using  the legacy journal DAO to the new journal DAO

I have tested the SchemaUtils with some PoC and it works like magic. I could not move data from the old journal to the new journal though. I did not get any error when running the application.

@octonato, @ennru  and @patriknw kindly take a look and advice.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ lazy val `akka-persistence-jdbc` = project
   .in(file("."))
   .enablePlugins(ScalaUnidocPlugin)
   .disablePlugins(MimaPlugin, SitePlugin)
-  .aggregate(core, migration, docs)
+  .aggregate(core, migration, docs, `schema-utils`)
   .settings(publish / skip := true)
 
 lazy val core = project
@@ -29,6 +29,16 @@ lazy val migration = project
     name := "akka-persistence-jdbc-migration",
     libraryDependencies ++= Dependencies.Migration,
     publish / skip := true)
+
+lazy val `schema-utils` = project
+  .in(file("schema-utils"))
+  .enablePlugins(Publish)
+  .disablePlugins(SitePlugin, MimaPlugin)
+  .settings(
+    name := "akka-persistence-jdbc-schema-utils",
+    libraryDependencies ++= Dependencies.Migration ++ Dependencies.Libraries,
+    publish / skip := true)
+  .dependsOn(core)
 
 lazy val docs = project
   .enablePlugins(ProjectAutoPlugin, AkkaParadoxPlugin, ParadoxSitePlugin, PreprocessPlugin, PublishRsyncPlugin)
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -33,7 +33,7 @@ object Dependencies {
     "org.scalatest" %% "scalatest" % ScalaTestVersion % Test) ++ JdbcDrivers.map(_ % Test)
 
   val Migration: Seq[ModuleID] = Seq(
-    "org.flywaydb" % "flyway-core" % "7.5.1",
+    "org.flywaydb" % "flyway-core" % "7.5.2",
     "com.typesafe" % "config" % "1.4.1",
     "ch.qos.logback" % "logback-classic" % "1.2.3",
     "org.testcontainers" % "postgresql" % "1.15.1" % Test,
--- a/schema-utils/src/main/resources/db/migration/postgres/V1__create-table.sql
+++ b/schema-utils/src/main/resources/db/migration/postgres/V1__create-table.sql
@@ -1,0 +1,56 @@
+-- creation of the new akka persistence jdbc schemas
+
+-- event_journal table
+CREATE TABLE IF NOT EXISTS ${apj:event_journal}
+(
+    ordering           BIGSERIAL,
+    persistence_id     VARCHAR(255)          NOT NULL,
+    sequence_number    BIGINT                NOT NULL,
+    deleted            BOOLEAN DEFAULT FALSE NOT NULL,
+
+    writer             VARCHAR(255)          NOT NULL,
+    write_timestamp    BIGINT,
+    adapter_manifest   VARCHAR(255),
+
+    event_ser_id       INTEGER               NOT NULL,
+    event_ser_manifest VARCHAR(255)          NOT NULL,
+    event_payload      BYTEA                 NOT NULL,
+
+    meta_ser_id        INTEGER,
+    meta_ser_manifest  VARCHAR(255),
+    meta_payload       BYTEA,
+
+    PRIMARY KEY (persistence_id, sequence_number)
+);
+
+CREATE UNIQUE INDEX ${apj:event_journal}_ordering_idx ON ${apj:event_journal} (ordering);
+
+-- event_tag table
+CREATE TABLE IF NOT EXISTS ${apj:event_tag}
+(
+    event_id BIGINT,
+    tag      VARCHAR(256),
+    PRIMARY KEY (event_id, tag),
+    CONSTRAINT fk_${apj:event_journal}
+        FOREIGN KEY (event_id)
+            REFERENCES ${apj:event_journal} (ordering)
+            ON DELETE CASCADE
+);
+
+-- state_snapshot table
+CREATE TABLE IF NOT EXISTS ${apj:state_snapshot}
+(
+    persistence_id        VARCHAR(255) NOT NULL,
+    sequence_number       BIGINT       NOT NULL,
+    created               BIGINT       NOT NULL,
+
+    snapshot_ser_id       INTEGER      NOT NULL,
+    snapshot_ser_manifest VARCHAR(255) NOT NULL,
+    snapshot_payload      BYTEA        NOT NULL,
+
+    meta_ser_id           INTEGER,
+    meta_ser_manifest     VARCHAR(255),
+    meta_payload          BYTEA,
+
+    PRIMARY KEY (persistence_id, sequence_number)
+);
--- a/schema-utils/src/main/scala/akka/persistence/jdbc/tools/LegacyJournalDataMigrator.scala
+++ b/schema-utils/src/main/scala/akka/persistence/jdbc/tools/LegacyJournalDataMigrator.scala
@@ -1,0 +1,62 @@
+package akka.persistence.jdbc.tools
+
+import akka.actor.ActorSystem
+import akka.persistence.journal.EventAdapter
+import akka.persistence.{ AtomicWrite, Persistence, PersistentRepr }
+import akka.stream.scaladsl.{ Sink, Source }
+import akka.NotUsed
+import com.typesafe.config.Config
+
+import scala.collection.immutable
+import scala.collection.immutable.Seq
+import scala.concurrent.Future
+import scala.util.Try
+
+/**
+ * helps migrate all events in the legacy journal to the new journal
+ *
+ * @param config
+ * @param system
+ */
+final case class LegacyJournalDataMigrator(config: Config)(implicit system: ActorSystem) {
+  import system.dispatcher
+
+  private val storesConfig = StoresConfig(config)
+  private val daos = MigrationDaos(storesConfig)
+
+  private val eventAdapters = Persistence(system).adaptersFor("", config)
+
+  private def adaptEvents(repr: PersistentRepr): Seq[PersistentRepr] = {
+    val adapter: EventAdapter = eventAdapters.get(repr.payload.getClass)
+    adapter.fromJournal(repr.payload, repr.manifest).events.map(repr.withPayload)
+  }
+
+  /**
+   * reads all the current events in the legacy journal
+   *
+   * @return the source of all the events
+   */
+  def readEvents(): Source[PersistentRepr, NotUsed] = {
+    daos.legacyReadJournalDao
+      .allPersistenceIdsSource(Long.MaxValue)
+      .flatMapConcat((persistenceId: String) => {
+        daos.legacyReadJournalDao
+          .messagesWithBatch(persistenceId, 0L, Long.MaxValue, storesConfig.readJournalConfig.maxBufferSize, None)
+          .mapAsync(1)(reprAndOrdNr => Future.fromTry(reprAndOrdNr))
+          .mapConcat { case (repr, _) =>
+            adaptEvents(repr)
+          }
+      })
+  }
+
+  /**
+   * write all legacy events into the new journal tables applying the proper serialization
+   */
+  def writeEvents(): Future[Future[Seq[Try[Unit]]]] = {
+    readEvents()
+      .runWith(Sink.seq) // TODO: find the best way to load all the events in memory
+      .map((sq: Seq[PersistentRepr]) => {
+        daos.defaultJournalDao.asyncWriteMessages(immutable.Seq(AtomicWrite(sq)))
+      })
+  }
+}
--- a/schema-utils/src/main/scala/akka/persistence/jdbc/tools/MigrationDaos.scala
+++ b/schema-utils/src/main/scala/akka/persistence/jdbc/tools/MigrationDaos.scala
@@ -1,0 +1,47 @@
+package akka.persistence.jdbc.tools
+
+import akka.actor.ActorSystem
+import akka.persistence.jdbc.db.SlickExtension
+import akka.persistence.jdbc.journal.dao.DefaultJournalDao
+import akka.persistence.jdbc.query.dao.legacy.ByteArrayReadJournalDao
+import akka.persistence.jdbc.snapshot.dao.DefaultSnapshotDao
+import akka.persistence.jdbc.snapshot.dao.legacy.ByteArraySnapshotDao
+import akka.serialization.SerializationExtension
+import slick.basic.DatabaseConfig
+import slick.jdbc
+import slick.jdbc.{ JdbcBackend, JdbcProfile }
+
+/**
+ * Wrapper class defining the various DAOs that are needed to run journal data migration
+ *
+ * @param storesConfig the stores config
+ * @param system the actor system
+ */
+final case class MigrationDaos(storesConfig: StoresConfig)(implicit system: ActorSystem) {
+  import system.dispatcher
+
+  // get an instance of the database and the Jdbc profile
+  val profile: JdbcProfile = DatabaseConfig.forConfig[JdbcProfile]("slick").profile
+
+  val journaldb: JdbcBackend.Database =
+    SlickExtension(system).database(storesConfig.config.getConfig("jdbc-read-journal")).database
+  val snapshotdb: jdbc.JdbcBackend.Database =
+    SlickExtension(system).database(storesConfig.config.getConfig("jdbc-snapshot-store")).database
+
+  // get an instance of the legacy read journal dao
+  val legacyReadJournalDao: ByteArrayReadJournalDao =
+    new ByteArrayReadJournalDao(journaldb, profile, storesConfig.readJournalConfig, SerializationExtension(system))
+
+  // get an instance of the default journal dao
+  val defaultJournalDao: DefaultJournalDao =
+    new DefaultJournalDao(journaldb, profile, storesConfig.journalConfig, SerializationExtension(system))
+
+  // get the instance of the legacy snapshot dao
+  val legacySnapshotDao: ByteArraySnapshotDao =
+    new ByteArraySnapshotDao(snapshotdb, profile, storesConfig.snapshotConfig, SerializationExtension(system))
+
+  // get the instance if the default snapshot dao
+  val defaultSnapshotDao: DefaultSnapshotDao =
+    new DefaultSnapshotDao(snapshotdb, profile, storesConfig.snapshotConfig, SerializationExtension(system))
+
+}
--- a/schema-utils/src/main/scala/akka/persistence/jdbc/tools/SchemasUtil.scala
+++ b/schema-utils/src/main/scala/akka/persistence/jdbc/tools/SchemasUtil.scala
@@ -1,0 +1,60 @@
+package akka.persistence.jdbc.tools
+
+import com.typesafe.config.Config
+import org.flywaydb.core.Flyway
+import org.flywaydb.core.api.Location
+import org.flywaydb.core.api.configuration.FluentConfiguration
+import org.flywaydb.core.api.output.MigrateResult
+
+import scala.jdk.CollectionConverters.{ CollectionHasAsScala, MapHasAsJava }
+
+/**
+ * executes the database migration
+ *
+ * @param config the application config
+ */
+final case class SchemasUtil(config: Config) {
+  private val userKey: String = "jdbc-journal.slick.db.user"
+  private val passwordKey: String = "jdbc-journal.slick.db.password"
+  private val urlKey: String = "jdbc-journal.slick.db.url"
+
+  // Flyway placeholders values keys
+  private val eventJournalPlaceholderValueKey: String = "jdbc-journal.tables.event_journal.tableName"
+  private val eventTagPlaceholderValueKey: String = "jdbc-journal.tables.event_tag.tableName"
+  private val stateSnapshotPlaceholderValueKey: String = "jdbc-snapshot-store.tables.snapshot.tableName"
+
+  // Flyway placeholders
+  private val eventJournalPlaceholder = "apj:event_journal"
+  private val eventTagPlaceholder: String = "apj:event_tag"
+  private val stateSnapshotPlaceholder: String = "apj:state_snapshot"
+
+  private val url: String = config.getString(urlKey)
+  private val user: String = config.getString(userKey)
+  private val password: String = config.getString(passwordKey)
+
+  /**
+   * create the persistence schemas.
+   *
+   * @return the list of migration versions run
+   */
+  def run(): Seq[String] = {
+    val flywayConfig: FluentConfiguration = Flyway.configure
+      .dataSource(url, user, password)
+      .table("apj_schema_history")
+      .locations(new Location("classpath:db/migration/postgres"))
+      .ignoreMissingMigrations(true) // in case someone has some missing migrations
+      .placeholders(Map(
+        eventJournalPlaceholder -> config.getString(eventJournalPlaceholderValueKey),
+        eventTagPlaceholder -> config.getString(eventTagPlaceholderValueKey),
+        stateSnapshotPlaceholder -> config.getString(stateSnapshotPlaceholderValueKey)).asJava)
+
+    val flyway: Flyway = flywayConfig.load
+    flyway.baseline()
+    // running the migration
+    val result: MigrateResult = flyway.migrate()
+
+    // let us return the sequence of migration versions applied sorted in a descending order
+    result.migrations.asScala.toSeq.map(_.version).sortWith(_ > _)
+  }
+
+}
--- a/schema-utils/src/main/scala/akka/persistence/jdbc/tools/StoresConfig.scala
+++ b/schema-utils/src/main/scala/akka/persistence/jdbc/tools/StoresConfig.scala
@@ -1,0 +1,10 @@
+package akka.persistence.jdbc.tools
+
+import akka.persistence.jdbc.config.{ JournalConfig, ReadJournalConfig, SnapshotConfig }
+import com.typesafe.config.Config
+
+final case class StoresConfig(config: Config) {
+  val journalConfig = new JournalConfig(config.getConfig("jdbc-journal"))
+  val snapshotConfig = new SnapshotConfig(config.getConfig("jdbc-snapshot-store"))
+  val readJournalConfig = new ReadJournalConfig(config.getConfig("jdbc-read-journal"))
+}


Tochemey:
@octonato , @patriknw for some weird reasons I am getting some type mismatch here https://travis-ci.com/github/akka/akka-persistence-jdbc/jobs/481103765#L368

build.sbt
```diff
@@ -30,6 +30,16 @@ lazy val migration = project
...
 
+lazy val tools = project
+  .in(file("tools"))
+  .enablePlugins(Publish)
+  .disablePlugins(SitePlugin, MimaPlugin)
+  .settings(
+    name := "akka-persistence-jdbc-schema-utils",
+    libraryDependencies ++= Dependencies.Migration ++ Dependencies.Libraries,
+    publish / skip := true)
+  .dependsOn(core)
```
octonato, line 41: I first thought that you were moving the SchemaUtils to that new artifact, but now I see that it's another SchemaUtils. Adapted for the migration. That's good. 👍 

We originally decided to keep the schema utils on the core package to avoid the overhead of having yet another package. 

Anyway, we don't need to have another module. The Flyway SchemaUtils can live inside the `migration` module.

tools/src/main/scala/akka/persistence/jdbc/tools/migration/LegacyJournalDataMigrator.scala
```diff
@@ -0,0 +1,61 @@
...
+ * @param system the actor system
+ */
+final case class LegacyJournalDataMigrator(config: Config)(implicit system: ActorSystem) extends DataMigrator(config) {
+
+  private val eventAdapters = Persistence(system).adaptersFor("", config)
+
+  private def adaptEvents(repr: PersistentRepr): Seq[PersistentRepr] = {
+    val adapter: EventAdapter = eventAdapters.get(repr.payload.getClass)
+    adapter.fromJournal(repr.payload, repr.manifest).events.map(repr.withPayload)
+  }
```
octonato, line 33: This is neat. 

This will allow users to eventually migrate their events to some new format.

There one thing that we should watch-out though. If an adapter produces more events then it takes (1 event split in many), that will change the ordering in the new table. 

That's not an issue in itself unless users are tracking the offset using Lagom or Akka Projections. 

That's all good, but we need to carefully inform users that their adapters should emit 1:1 in that case or they should reset their offset table and consume from scratch. 

There is one more thing on the adapters approach, it that we can re-tag the events.

tools/src/main/scala/akka/persistence/jdbc/tools/migration/LegacyJournalDataMigrator.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * reads all the current events in the legacy journal
+   *
+   * @return the source of all the events
+   */
+  private def allEvents(): Source[PersistentRepr, NotUsed] = {
+    legacyReadJournalDao
+      .allPersistenceIdsSource(Long.MaxValue)
+      .flatMapConcat((persistenceId: String) => {
```
octonato, line 43: The mapConcat will change the ordering in the new schema. Instead, we should read the events directly without fetching by persistence id.

tools/src/main/scala/akka/persistence/jdbc/tools/migration/LegacySnapshotDataMigrator.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  import system.dispatcher
+
+  /**
+   * write the latest state snapshot into the new snapshot table applying the proper serialization
+   */
+  def migrate(): Source[Option[Future[Unit]], NotUsed] = {
+    legacyReadJournalDao.allPersistenceIdsSource(Long.MaxValue).mapAsync(1) { persistenceId: String =>
+      legacySnapshotDao
+        .latestSnapshot(persistenceId)
```
octonato, line 33: We may want to offer the option to migrate the last `N` snapshots. Is common for users to keep more than snapshot and in that case they probably want to migrate them as well.

octonato (commented):
@Tochemey, this is looking good. 

I like the event adapter idea. 

I'm wondering if we should take a similar path as the Cassandra plugin. The Cassandra has a Cleanup tool with different methods that users can combine to achieve what they want. 

For instance, do I need to preserve the ordering of events? I may want to do it if I'm using Akka Projections and don't want to replay all my projections. 

In that case, preserving the ordering is a must. 

If I don't care about the ordering, I may even migrate per persistence id and in parallel.

octonato:
@Tochemey, I have a second thought about the Flyway part. Maybe we should just drop it. There is too much on our side to have something that would work for all databases. Selecting the type, etc. At the end, what is the real added value? 

Maybe the best is to let users just pick the schema as we publish in the docs (or use the new SchemaUtils) to create it. Then run the tool using new legacy and new DAOs. And that's it. 

The migration doesn't need to take care of creating the schema for the users.

Tochemey:
However the new SchemaUtils need to be made a bit flexible to accommodate the user defined postgres schema. At the moment it is tied to the `public` schema which can be problematic.

tools/src/main/scala/akka/persistence/jdbc/tools/migration/LegacyJournalDataMigrator.scala
Tochemey: At the moment there is no utility function to help read all messages in the journal. We may have to resort to some direct slick calls to fetch all data in the journal table which is not a big deal.

tools/src/main/scala/akka/persistence/jdbc/tools/migration/LegacyJournalDataMigrator.scala
Tochemey: Interesting.

build.sbt
Tochemey: I have provided a fix in the latest commits.

Tochemey:
@octonato I was able to migrate old journal and snapshot at a very scale. However I observed that the event tags have not been migrated at all. Is there anything I have to do to get them migrated while migrating the journal data.

Tochemey:
@patriknw some lovely feedback will not be bad.😁

Tochemey:
@octonato and @patriknw I am able to migrate all data from snapshot and journal into the new tables. However for the tags migration this is the code I use. 

```scala
private def events(): Source[PersistentRepr, NotUsed] = {
    Source
      .fromPublisher(
        journaldb.stream(queries.JournalTable.sortBy(_.sequenceNumber).result)
      )
      .via(serializer.deserializeFlow)
      .mapAsync(1)((reprAndOrdNr: Try[(PersistentRepr, Set[String], Long)]) => Future.fromTry(reprAndOrdNr))
      .map { case (repr, tags, _) =>
        repr.withPayload(Tagged(repr.payload, tags))
      }
  }
```

I don't think we will need an event adapter since the payload is the raw one that is persisted into the old journal table column as well as the tags. I strongly believe we just need to move the data.

octonato:
> I don't think we will need an event adapter since the payload is the raw one that is persisted into the old journal table column as well as the tags. I strongly believe we just need to move the data.

@Tochemey, I think you are right on this. 

The good part of this approach is that it doesn't leave the door open for adding more events. An event adapter is capable of taking one event and emit more than one. That would add more events on the new table.

build.sbt
```diff
@@ -4,7 +4,7 @@ lazy val `akka-persistence-jdbc` = project
   .in(file("."))
   .enablePlugins(ScalaUnidocPlugin)
   .disablePlugins(MimaPlugin, SitePlugin)
-  .aggregate(core, migration, docs)
+  .aggregate(core, migration, docs, migrator)
```
octonato, line 7: Let's remove the `migration` project and keep only the `migrator`.

build.sbt
```diff
@@ -30,6 +30,16 @@ lazy val migration = project
...
     publish / skip := true)
 
+lazy val migrator = project
+  .in(file("migrator"))
+  .enablePlugins(Publish)
+  .disablePlugins(SitePlugin, MimaPlugin)
+  .settings(
+    name := "akka-persistence-jdbc-migrator",
+    libraryDependencies ++= Dependencies.Migration ++ Dependencies.Libraries,
+    publish / skip := true)
```
octonato, line 40: Add a TODO here so we don't forget to remove it when ready.

We want to publish the tool, but only when ready for use. It doesn't need to be when production ready. We may decide to publish it on a snapshot to test it out on real world projects.

migrator/src/main/scala/akka/persistence/jdbc/migrator/LegacyJournalDataMigrator.scala
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /**
+   * write all legacy events into the new journal tables applying the proper serialization
+   */
+  def migrate(): Unit = {
+    events()
+      .mapAsync(1)((repr: PersistentRepr) => {
+        defaultJournalDao.asyncWriteMessages(Seq(AtomicWrite(Seq(repr))))
+      })
```
octonato, line 80:
```suggestion
      .mapAsync(1) { repr => 
        defaultJournalDao.asyncWriteMessages(Seq(AtomicWrite(Seq(repr))))
      }
```

project/Dependencies.scala
```diff
@@ -33,7 +33,7 @@ object Dependencies {
     "org.scalatest" %% "scalatest" % ScalaTestVersion % Test) ++ JdbcDrivers.map(_ % Test)
 
   val Migration: Seq[ModuleID] = Seq(
-    "org.flywaydb" % "flyway-core" % "7.5.1",
+    "org.flywaydb" % "flyway-core" % "7.5.2",
```
octonato, line 36: Lets' remove this. We don't need it.

migrator/src/main/scala/akka/persistence/jdbc/migrator/LegacyJournalDataMigrator.scala
```diff
@@ -0,0 +1,85 @@
...
+      .limit(Long.MaxValue)
+      .runWith(Sink.seq) // FIXME for performance
```
octonato, line 82: `Sing.ignore` is good enough here and we won't accumulate records in memory.

migrator/src/main/scala/akka/persistence/jdbc/migrator/LegacyJournalDataMigrator.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * write all legacy events into the new journal tables applying the proper serialization
+   */
+  def migrate(): Unit = {
+    events()
+      .mapAsync(1)((repr: PersistentRepr) => {
+        defaultJournalDao.asyncWriteMessages(Seq(AtomicWrite(Seq(repr))))
+      })
+      .limit(Long.MaxValue)
```
octonato, line 81: I don't think we need it. The slick query should already provide a stream that is only consumed as we go. 

We may need to double check how many records it's fetching each time, but I'm sure it's good enough.

migrator/src/main/scala/akka/persistence/jdbc/migrator/LegacyJournalDataMigrator.scala
```diff
@@ -0,0 +1,85 @@
...
+   * @return
+   */
+  private def events(): Source[PersistentRepr, NotUsed] = {
+    Source
+      .fromPublisher(journaldb.stream(journalQueries.JournalTable.sortBy(_.sequenceNumber).result))
+      .via(serializer.deserializeFlow)
+      .mapAsync(1)((reprAndOrdNr: Try[(PersistentRepr, Set[String], Long)]) => Future.fromTry(reprAndOrdNr))
+      .map { case (repr, tags, _) =>
+        repr.withPayload(Tagged(repr.payload, tags))
+      }
```
octonato, line 70: We don't need to go from `Try` to `Future` and back to unwrapped tuple 3. 

I think that whole block can be changed to: 

```scala
.map {
  case Success((repr, tags, _)) if tags.nonEmpty => 
    repr.withPayload(Tagged(repr, tags))      // only wrap in `Tagged` if needed
  case Success((repr, _, _)) => repr          // noops map
  case Failure(exception) => throw exception  // blow-up on failure
}
```
Eventually we need to add more logging inside `deserializeFlow`. If something goes wrong, that will be when deserializing the payload.

migrator/src/main/scala/akka/persistence/jdbc/migrator/LegacySnapshotDataMigrator.scala
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * migrate all the legacy snapshot schema data into the new snapshot schema
+   */
+  def migrateAll(): Future[Seq[Future[Unit]]] = {
+    for {
+      rows <- snapshotdb.run(queries.SnapshotTable.sortBy(_.sequenceNumber.desc).result)
+    } yield rows.map(toSnapshotData).map { case (metadata, value) =>
+      defaultSnapshotDao.save(metadata, value)
+    }
+  }
```
octonato, line 68: Snapshots tend to have payloads much larger then events. I think this query will fetch all snapshots into memory. Better if we stream like we did for the events.

migrator/src/main/scala/akka/persistence/jdbc/migrator/LegacySnapshotDataMigrator.scala
```diff
@@ -0,0 +1,85 @@
...
+    for {
+      rows <- snapshotdb.run(queries.SnapshotTable.sortBy(_.sequenceNumber.desc).drop(offset).take(limit).result)
+    } yield rows.map(toSnapshotData).map { case (metadata, value) =>
+      defaultSnapshotDao.save(metadata, value)
+    }
+  }
+
+  def migrateLatest(): Future[Option[Future[Unit]]] = {
+    for {
+      rows <- snapshotdb.run(queries.SnapshotTable.sortBy(_.sequenceNumber.desc).take(1).result)
```
octonato, line 80: This will only take the latest snapshot across all persistence ids. 

Insteads, we should use get the distinct list of persistence id (in a stream), run through it and fetch the latest per persistenceId.

migrator/src/main/scala/akka/persistence/jdbc/migrator/LegacySnapshotDataMigrator.scala
```diff
@@ -0,0 +1,85 @@
...
+  private def toSnapshotData(row: SnapshotRow): (SnapshotMetadata, Any) =
+    serializer.deserialize(row) match {
+      case Success(deserialized) => deserialized
+      case Failure(cause)        => throw cause
+    }
+
+  /**
+   * migrate all the legacy snapshot schema data into the new snapshot schema
+   */
+  def migrateAll(): Future[Seq[Future[Unit]]] = {
```
octonato, line 62: `Future[Seq[Future[Unit]]]` this will confuse users. When is the migration done? When the outer future completes I still have a bunch on Futures in flight. 

Better if we return a `Future[Done]` here and that we do it using a stream of snapshots. We consume one by one and we finish with `Sink.ignore` and return a `Future[Done]`.

migrator/src/main/scala/akka/persistence/jdbc/migrator/LegacySnapshotDataMigrator.scala
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }
+
+  def migrate(offset: Int, limit: Int): Future[Seq[Future[Unit]]] = {
+    for {
+      rows <- snapshotdb.run(queries.SnapshotTable.sortBy(_.sequenceNumber.desc).drop(offset).take(limit).result)
+    } yield rows.map(toSnapshotData).map { case (metadata, value) =>
+      defaultSnapshotDao.save(metadata, value)
+    }
+  }
```
octonato, line 76: Maybe this one is less useful. I see the value in migrating only the latest snapshot like in the following method. But migrating a subset is maybe something that we can skip. 

It's also one scenario less to test for us.

octonato (commented):
Thanks @Tochemey. 

I think we are going in a good direction. Cool that you manage to read the tags and remove the event adapter. 

We will need to think of integration tests. 
First start the system with old dao, persist some events and snapshots. Stop it, migrate the data and start again using the new DAOs.

migrator/src/main/scala/akka/persistence/jdbc/migrator/JournalMigrator.scala
```diff
@@ -0,0 +1,177 @@
...
+      .map {
+        case Success((repr, tags, ordering)) if tags.nonEmpty =>
+          repr.withPayload(Tagged(repr, tags)) -> ordering // only wrap in `Tagged` if needed
+        case Success((repr, _, ordering)) => repr -> ordering // noops map
+        case Failure(exception)           => throw exception // blow-up on failure
+      }
+      // since the write is done one at a time we can at least enhance throughput by
+      // spawning more actors to execute the db write
+      // FIXME we can figure out a way to do some batch inserts because the ordering of insertion does not matter any longer
+      .mapAsync(parallelism) { case (repr, ordering) =>
```
octonato, line 93: although we are preserving the ordering, I would just insert in the order we read and then set this to `mapAsync(1)`. 

The natural order for many DBs, if I'm not mistaking, is the order that bytes are saved on disk. If we use `mapAsync(8)` we will interleave then on disk. We don't need to do it.

For the record, I create a journal with 500k events, each events with 200 bytes of payload. The migration took around 20 minutes with `mapAsync(1)`.

migrator/src/main/scala/akka/persistence/jdbc/migrator/SnapshotMigrator.scala
```diff
@@ -0,0 +1,99 @@
...
+ * appropriate serialization
+ *
+ * @param system the actor system
+ */
+case class SnapshotMigrator()(implicit system: ActorSystem) {
+
+  import system.dispatcher
+
+  // get the Jdbc Profile
+  protected val profile: JdbcProfile = DatabaseConfig.forConfig[JdbcProfile]("slick").profile
```
octonato, line 37: here the same, we can use `DatabaseType`

migrator/src/main/scala/akka/persistence/jdbc/migrator/SnapshotMigrator.scala
```diff
@@ -0,0 +1,99 @@
...
+   */
+  def migrateAll(): Future[Done] = {
+    Source
+      .fromPublisher(snapshotdb.stream(queries.SnapshotTable.result))
+      .map((record: SnapshotRow) => {
+        val (metadata, value) = toSnapshotData(record)
+        defaultSnapshotDao.save(metadata, value)
+      })
+      .runWith(Sink.ignore)
+  }
```
octonato, line 98: I need to check further in the code, but this is not working as expected. 

In my trial, I had 10000 snapshots for 5000 different entities (2 per entity). I logged them and I can see that all 10000 snapshots are printed. However, I don't get then all in the database. 

I lost quite a lot. It inserted less than 1000 snapshots.

migrator/src/main/scala/akka/persistence/jdbc/migrator/package.scala
```diff
@@ -0,0 +1,142 @@
...
+  case object H2 extends DatabaseType
+
+  case object Oracle extends DatabaseType
+
+  /**
+   * helps set the new journal ordering value after the data migration is done.
+   * This trait will be implemented by the various rdbms supported by the akka persistence jdbc since
+   * the sequence number is handled differently by the various databases
+   */
+  sealed trait JournalOrdering {
```
octonato, line 33: @Tochemey, I still need to wrap my head around this. I did tests today without looking further on the mechanism here. 

Could you explain why we need it?

migrator/src/main/scala/akka/persistence/jdbc/migrator/SnapshotMigrator.scala
```diff
@@ -0,0 +1,99 @@
...
+   */
+  def migrateLatest(): Future[Done] = {
+    legacyJournalDao
+      .allPersistenceIdsSource(Long.MaxValue)
+      .map(persistenceId => {
+        // let us fetch the latest snapshot for each persistenceId
+        snapshotdb
+          .run(queries.selectLatestByPersistenceId(persistenceId).result)
+          .map(rows => {
+            rows.headOption.map(toSnapshotData).map { case (metadata, value) =>
```
octonato, line 79: also good to add debug logs here

migrator/src/main/scala/akka/persistence/jdbc/migrator/SnapshotMigrator.scala
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  /**
+   * migrate all the legacy snapshot schema data into the new snapshot schema
+   */
+  def migrateAll(): Future[Done] = {
+    Source
+      .fromPublisher(snapshotdb.stream(queries.SnapshotTable.result))
+      .map((record: SnapshotRow) => {
+        val (metadata, value) = toSnapshotData(record)
```
octonato, line 94: logs here as well

octonato (commented):
@Tochemey, I took some time today to play around with the tool. It's looking great. 

I migrated 500k journal without major problems. 

I left a few comments explaining the issues I found. 

One thing that I didn't realised before is that the new schema use the same table name for the snapshot table. 

So, I had to first rename my snapshot table to `snapshot_old`, set the legacy config to read from it and run the migration. 

I will run some more tests and report here.

migrator/src/main/scala/akka/persistence/jdbc/migrator/SnapshotMigrator.scala
octonato: Ok, I found it. 

We need to use `mapAsync(1)` instead of `map`.

octonato: Bingo!!

```
docker=# select count(*) from snapshot_old;
 count
-------
 10000
(1 row)
```

```
docker=# select count(*) from snapshot;
 count
-------
 10000
(1 row)
```

migrator/src/main/scala/akka/persistence/jdbc/migrator/JournalMigrator.scala
```diff
@@ -0,0 +1,177 @@
...
+    val journalInsert: DBIO[Long] = newJournalQueries.JournalTable
+      .returning(newJournalQueries.JournalTable.map(_.ordering))
+      .forceInsert(
+        journalSerializedRow
+      ) // here we force the insertion of the auto-inc value to maintain the old ordering
+
+    val tagInserts: FixedSqlAction[Option[Int], NoStream, Effect.Write] =
+      newJournalQueries.TagTable ++= tags.map(tag => TagRow(journalSerializedRow.ordering, tag)).toSeq
+
+    journaldb.run(DBIO.seq(journalInsert, tagInserts).withPinnedSession.transactionally)
```
octonato, line 175: have you tried to do the following? 

```scala
val combined = journalInsert.flatMap(_ => tagInserts)
journaldb.run(combined.transactionally)
```
I think the difference is that with `seq` one statement is executed against the DB and then other (two round trips). While with `flatMap` they are both sent to the DB as one command. But I'm not 100% sure about it. 

Also, we may not need a pinned session.

octonato:
@Tochemey about adding tests...

Our test infrastructure is as such:

* we have an abstract class that define the whole test
* then we have H2 variant run as 'unit test' and docker variants run as 'integration tests'

That pattern repeats all over the tests. Most test start one actor system per suite using pre-defined .conf file. Some tests start a actor system per test.

For the migration tests we will need something different. 

We will need to start and stop two different actor system because they need to load different config. 

So for instance, 

```scala
it should "migrate the journal" in {
  withActorSystem(configWithLegacyDB) { sys => 
	// persist some events

	// migrate events (can use same actor system)
  }

  // start new sys
  withActorSystem(configWithNewDB) { sys => 
	// check can load events

    // check eventsByTag to verify tags were migrated
  }
}

```

You can start looking that this test for example:
https://github.com/akka/akka-persistence-jdbc/blob/95bb47e7196d60a6f76a62ad17e2ae89066a01d7/core/src/test/scala/akka/persistence/jdbc/query/AllPersistenceIdsTest.scala#L11 for inspiration. This is for the queries, but we can consider something similar.

migrator/src/main/scala/akka/persistence/jdbc/migrator/JournalMigrator.scala
Tochemey: Ok we can also group them and do some batch insertions.

migrator/src/main/scala/akka/persistence/jdbc/migrator/package.scala
Tochemey: The goal is to help after the migration to reset the event_journal ordering to start with where it was in the old journal.

Tochemey:
@octonato I think for the tests we may only to find a way to load data into the legacy tables and migrating that data into the new journal. A single actor system should suffice per various tests according to the databases supported.

migrator/src/main/scala/akka/persistence/jdbc/migrator/package.scala
octonato: Oh! I see it now.

So, the basic issue is that we are forcing new ordering because we are importing the ordering from another table and therefore, when we finished, the new table sequence/auto-increment needs to be reconfigured to start from the last + 1. Correct? 

I remember that at some point you mentioned that you were getting the ordering interleaved, but I think the main reason was because when streaming the results you were calling `sortBy(_.sequenceNumber)`.

Wouldn't we just get the right ordering if we read on one side, respecting the ordering, and insert on the other? At the end of the game, the ordering on the new table should be exactly the same. Or am I missing something obvious? 

The `JournalOrdering` you introduced looks neat if we don't have another solution. But I would first try to just ensure that the ordering is preserved by inserting on the same order. As such we will have less code.

migrator/src/main/scala/akka/persistence/jdbc/migrator/package.scala
```diff
@@ -0,0 +1,162 @@
...
+  JdbcProfile,
+  MySQLProfile,
+  OracleProfile,
+  PostgresProfile,
+  SQLServerProfile
+}
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+package object migrator {
```
octonato, line 22: I don't think we need to add a package object. We can just add a `DatabaseType.scala` and `JournalOrdering.scala` file. 

It makes it easier to find them when just browsing the code in GitHub, text editor or file browser. When we add traits and classes to a package.scala, we need to know they live there in order to find them.

migrator/src/test/scala/akka/persistence/jdbc/migrator/BaseJournalMigratorSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+import akka.persistence.jdbc.journal.dao.legacy
+import akka.persistence.PersistentRepr
+import akka.serialization.Serialization
+import akka.Done
+import slick.jdbc.JdbcBackend
+
+import java.util.UUID
+import scala.concurrent.{ ExecutionContext, Future }
+
+abstract class BaseJournalMigratorSpec(config: String) extends SingleActorSystemPerTestSpec(config) {
```
octonato, line 18: I think it will be more convenient (and closer to user experience) if the tests start a `EventSourcedBehavior` using the legacy tables and fill it by sending commands instead of going low-level and manipulate the queries directly. 

When doing so, we can also configure the behavior to save the snapshots every N events (eg: N=5) and to keep 3 snapshots for example. 

That's why I was saying that we may need to start two actor systems. 

I suggest the following test plan: 

1. start actor system with the legacy config and populate the journal for a few entities, with many events and snapshots. Stop actor system
2. run migrator for events and snapshots
3. start another actor system using default config. Reload each entity and verify that their state is as expected. 

The Account entity is a good one because we can easily test the balance.

Another tests could be: 

1. start actor system with the legacy config and populate the journal for a few entities, with many events and **save snapshot on each event**. Stop actor system
2. run migrator only for snapshots
3. start another actor system using default config. Reload each entity and verify that they state is as expected. 

In that second test, since we save snapshots on each event we don't need to migrate the events and when verifying the state, we prove that we indeed migrated all snapshots.

Finally, we need to test that the tags are correctly migrated. 
For that test I think it suffice to create one entity that persist events with tags, migrate the events and consume the tags using `eventsByTags` from the new table. 

In that test, we can simply verify that we consume the same amount of events as we generated through the entity. 

WDYT?

octonato (commented):
@Tochemey, I left a few more comments and a proposal for a test plan.

migrator/src/main/scala/akka/persistence/jdbc/migrator/package.scala
Tochemey: The ordering can never be respected because it is generated by the database. So the only way to ensure the ordering from the old journal to the new is to force insert them during the insertion into the new journal because the ordering field is a database auto-increment field. So after the migration of the data reset the ordering per database on the new journal as done here.

octonato: but why wouldn't the ordering be respected if the records are inserted in the same order? True that the ordering is generated by the db, but it is deterministic, no? 

If we have a blank slate and we insert the first event, the ordering will be 1, second event ordering 2, etc. 

What am I missing?

flonedo:
Hello, is this still being worked on? We have some production code we would like to upgrade to a newer akka-persistence version, and this PR would be extremely useful for us. We'd be happy to contribute with testing if necessary. Thanks in advance.

gabriele83:
We'd be happy to help, there is a way we can contribute?

migrator/src/main/scala/akka/persistence/jdbc/migrator/JournalMigrator.scala
```diff
@@ -0,0 +1,187 @@
...
+    } yield Done
+  }
+
+  /**
+   * Unpack a PersistentRepr into a PersistentRepr and set of tags.
+   * It returns a tuple containing the PersistentRepr and its set of tags
+   *
+   * @param pr the given PersistentRepr
+   */
+  private def unpackPersistentRepr(pr: PersistentRepr): (PersistentRepr, Set[String]) = {
```
gabriele83, line 130: never used declaration

migrator/src/main/scala/akka/persistence/jdbc/migrator/JournalMigrator.scala
```diff
@@ -0,0 +1,187 @@
...
+   * @param pr       the PersistentRepr
+   * @param ordering the ordering of the PersistentRepr
+   * @return the tuple of JournalAkkaSerializationRow and set of tags
+   */
+  private def serialize(
+      repr: PersistentRepr,
+      tags: Set[String],
+      ordering: Long): (JournalAkkaSerializationRow, Set[String]) = {
+
+    val serializedPayload: AkkaSerialization.AkkaSerialized =
```
gabriele83, line 149: val serializedPayload: AkkaSerialization.AkkaSerialized =
      AkkaSerialization.serialize(serialization, repr.payload).get

migrator/src/main/scala/akka/persistence/jdbc/migrator/JournalOrdering.scala
```diff
@@ -0,0 +1,121 @@
...
+
+sealed trait JournalOrdering {
+  val journalConfig: JournalConfig
+  val journalQueries: JournalQueries
+  val database: JdbcBackend.Database
+
+  /**
+   * the actual event journal table
+   */
+  val tableName: String = s"${journalConfig.eventJournalTableConfiguration.tableName}"
```
gabriele83, line 22: it is enough only:

`val tableName: String = journalConfig.eventJournalTableConfiguration.tableName`

migrator/src/main/scala/akka/persistence/jdbc/migrator/SnapshotMigrator.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  // get the instance if the default snapshot dao
+  private val defaultSnapshotDao: DefaultSnapshotDao =
+    new DefaultSnapshotDao(snapshotdb, profile, snapshotConfig, serialization)
+
+  // get the instance of the legacy journal DAO
+  private val legacyJournalDao: ByteArrayReadJournalDao =
+    new ByteArrayReadJournalDao(journaldb, profile, readJournalConfig, SerializationExtension(system))
+
+  private def toSnapshotData(row: SnapshotRow): (SnapshotMetadata, Any) =
```
gabriele83, line 60: private def toSnapshotData(row: SnapshotRow): (SnapshotMetadata, Any) =
    serializer.deserialize(row).get

octonato:
@flonedo / @gabriele83, I think the best way to contribute on this PR is to see with @Tochemey as he's the original author. 

There are a few paths we can follow: 

1. @Tochemey, gives access to his repo to both of you and you push code to the branch used in this PR (complicated solution)
2. you fork @Tochemey fork and send PRs to his fork targeting the branch used here. Once merged there, the changes will show up on this PR (also complicated). 
3. I merge this branch as is, but not on master, but on another branch. You guys can continue work sending PRs to this repo, but targeting the new 'feature' branch. Once the branch is ready, we merge it into master. 

I think option 3 is the easiest and will allow others to contribute, but first we need to know from @Tochemey if he will continue work on this.

gabriele83:
@octonato I have already opened a PR 4 days ago on the fork of @Tochemey, I'm waiting ...

Tochemey:
@gabriele83 The migration code has been used in production already and it was successful(we were only using postgres and we did get all relevant tests). However because of the general nature of the lib there might be some edge cases. Honestly I believe it should work for any of the because we migrate the raw data from the old schema unto the new. The only thing that remains that I can think of at the moment are the tests(unit and integration) which I have not got the time to write. Also @octonato will be the best person to give us the edge cases that can be an issue. 

I am glad to assist whenever necessary.

Tochemey:
@gabriele83 I have left a comment on you PR.

Tochemey:
@gabriele83 I have added you as collab on my branch. When free kindly fix the conflicts issue and let us merge it to this PR.

gabriele83:
Ok 👍

Issue #504, opened 2021-02-10
Missing breaking documentation for 4.X to 5.X migration ?


rdubois:
### Versions used 

Akka version: 2.6.12

### Expected Behavior

The migration documentation is about changing the DAOs to be used when using old schemas. I think there is a missing part in the documentation related to `JdbcAsyncWriteJournal` which relies on `Legacy` table configuration to create the journal config.

Before the change, the `JdbcAsyncWriteJournal` did use that kind of config:

```
jdbc-journal {
  class = "akka.persistence.jdbc.journal.JdbcAsyncWriteJournal"

  tables {
    journal {
     tableName = "journal"
     schemaName = ""
   }
  }
}
```

It seems it now must be changed to:

```
jdbc-journal {
  class = "akka.persistence.jdbc.journal.JdbcAsyncWriteJournal"

  tables {
    legacy_journal {
     tableName = "journal"
     schemaName = ""
   }
  }
}
```

And this is the same for the snapshot table.

Did I miss anything ?

### Actual Behavior

N/A

### Relevant logs

N/A

### Reproducible Test Case

N/A

octonato:
@rdubois, you are right. The config moved to `legacy_journal`. Same for the snapshot.

Thanks for reporting it.

rdubois:
Documentation has been updated. I clode that issue.

PR #506, opened 2021-02-11
Mention change in configuration key for legacy schema

octonato:
References #504

### DIFF
--- a/docs/src/main/paradox/migration.md
+++ b/docs/src/main/paradox/migration.md
@@ -4,7 +4,7 @@
 
 **Release `5.0.0` introduces a new schema and serialization that is not compatible with older versions.** 
 
-The previous version was wrapping the event payload with Akka's `PersistentRepr`, while in 5.0.0 the serialized event payload is persisted directly into the column. In order to migrate to the new schema, a migration tool capable of reading the serialized representation of `PersistentRepr` is required. That tool doesn't exist yet, therefore, the new schema can only be used with new applications.
+The previous version was wrapping the event payload with Akka's `PersistentRepr`, while in 5.0.0 the serialized event payload is persisted directly into the column. In order to migrate to the new schema, a migration tool capable of reading the serialized representation of `PersistentRepr` is required. That [tool doesn't exist yet](https://github.com/akka/akka-persistence-jdbc/issues/317), therefore, the new schema can only be used with new applications.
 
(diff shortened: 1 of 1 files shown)


docs/src/main/paradox/migration.md
```diff
@@ -23,3 +23,8 @@ jdbc-read-journal {
   dao = "akka.persistence.jdbc.query.dao.legacy.ByteArrayReadJournalDao"
 }
 ```
+
+If you have re-configured the `schemaName`, `tableName` and `columnNames` through configuration settings, you will need to move them a new key.
+
+* key `jdbc-journal.tables.journal` becomes `jdbc-journal.tables.legacy_journal`
+* key `jdbc-journal.tables.snapshot` becomes `jdbc-journal.tables.legacy_snapshot`
```
rdubois, line 30:
```suggestion
* key `jdbc-snapshot-store.tables.snapshot` becomes `jdbc-snapshot-store.tables.legacy_snapshot`
* key `jdbc-read-journal.tables.journal` becomes `jdbc-read-journal.tables.legacy_journal`
```

rdubois: @octonato here is a proposal to be even more exhaustive with the add of the `jdbc-read-journal` part. Also, I guess the change regarding the snapshot tables is about the `jdbc-snapshot-store` key.

octonato: You are right, that was a mistake of mine. 

For the read journal is less necessary. The old dao is now reading `jdbc-read-journal.tables.legacy_journal` and the reference.conf takes the value from `jdbc-journal.tables.legacy_journal`, so at the end it will already point to the right place.

Anyway, it does no harm to mention it. Thanks

docs/src/main/paradox/migration.md
```diff
@@ -23,3 +23,9 @@ jdbc-read-journal {
   dao = "akka.persistence.jdbc.query.dao.legacy.ByteArrayReadJournalDao"
 }
 ```
+
+If you have re-configured the `schemaName`, `tableName` and `columnNames` through configuration settings, you will need to move them a new key.
```
seglo, line 27:
```suggestion
If you have re-configured the `schemaName`, `tableName` and `columnNames` through configuration settings then you will need to move them to a new key.
```

seglo (approved):
LGTM. One docs suggestion.

PR #507, opened 2021-02-12
Remove travis and fix credentials

octonato:


.github/workflows/release.yml
```diff
@@ -42,15 +42,18 @@ jobs:
...
 
       - name: Publish API and reference documentation
         env:
           GUSTAV_KEY: ${{ secrets.GUSTAV_KEY }}
+          GUSTAV_PASSPHRASE: ${{ secrets.GUSTAV_PASSPHRASE }}
         run: |+
           eval "$(ssh-agent -s)"
           echo $GUSTAV_KEY | base64 -di > .github/id_rsa
           chmod 600 .github/id_rsa
+          ssh-keygen -p -P "$GUSTAV_PASSPHRASE" -N "" -f .github/id_rsa
```
octonato, line 57: In travis we were uploading the private key and then using a passphrase to open it. So, I'm reproducing the same here except that the key is now stored in a secret. 

It would be better to generate a new key since the previous were published here. The passphase was never compromised tough. Still, better safe than sorry.

octonato:
I tried to disable traivs in the settings, the it was triggered anyway. 🤷

octonato:
@chbatey since you mentioned you were planning some gardening  on the JDBC plugin. Here is a quick one that is just waiting for a ✅

chbatey (approved):

PR #533, opened 2021-05-21
Fix performance of _highestSequenceNrForPersistenceId slick query

jaceksokol:
There is a problem with the performance of `_highestSequenceNrForPersistenceId` query (generated by slick).
It's causing a long recovery time on my dataset (~ 69 millions of records in the journal). Adding the limit changes the generated query and drastically improves performance (in my case it was reduces from seconds to milliseconds).

Previous query:
`select max(x2.x3) from (select "persistence_id" as x4, "sequence_number" as x3 from "event_journal" where "persistence_id" = '[id]' order by "sequence_number" desc) x2;`
```
Aggregate  (cost=8524.87..8524.88 rows=1 width=8)
  ->  Index Only Scan Backward using event_journal_pkey on event_journal  (cost=0.57..8430.38 rows=7559 width=524)
        Index Cond: (persistence_id = '[id]'::text)
```

After the change:
`explain select max(x2.x3) from (select "persistence_id" as x4, "sequence_number" as x3 from "event_journal" where "persistence_id" = '[id]' order by "sequence_number" desc limit 1) x2;`
```
Aggregate  (cost=1.70..1.71 rows=1 width=8)
  ->  Limit  (cost=0.57..1.68 rows=1 width=524)
        ->  Index Only Scan Backward using event_journal_pkey on event_journal  (cost=0.57..8430.38 rows=7559 width=524)
              Index Cond: (persistence_id = '[id]'::text)
```

### DIFF
--- a/core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
@@ -52,10 +52,10 @@ class JournalQueries(
       .update(true)
 
   private def _highestSequenceNrForPersistenceId(persistenceId: Rep[String]): Rep[Option[Long]] =
-    selectAllJournalForPersistenceId(persistenceId).map(_.sequenceNumber).max
+    selectAllJournalForPersistenceId(persistenceId).take(1).map(_.sequenceNumber).max
 
   private def _highestMarkedSequenceNrForPersistenceId(persistenceId: Rep[String]): Rep[Option[Long]] =
-    selectAllJournalForPersistenceId(persistenceId).filter(_.deleted === true).map(_.sequenceNumber).max
+    selectAllJournalForPersistenceId(persistenceId).filter(_.deleted === true).take(1).map(_.sequenceNumber).max
 
   val highestSequenceNrForPersistenceId = Compiled(_highestSequenceNrForPersistenceId _)
 
--- a/core/src/main/scala/akka/persistence/jdbc/journal/dao/legacy/JournalQueries.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/journal/dao/legacy/JournalQueries.scala
@@ -47,10 +47,10 @@ class JournalQueries(val profile: JdbcProfile, override val journalTableCfg: Leg
       .update(true)
 
   private def _highestSequenceNrForPersistenceId(persistenceId: Rep[String]): Rep[Option[Long]] =
-    selectAllJournalForPersistenceId(persistenceId).map(_.sequenceNumber).max
+    selectAllJournalForPersistenceId(persistenceId).take(1).map(_.sequenceNumber).max
 
   private def _highestMarkedSequenceNrForPersistenceId(persistenceId: Rep[String]): Rep[Option[Long]] =
-    selectAllJournalForPersistenceId(persistenceId).filter(_.deleted === true).map(_.sequenceNumber).max
+    selectAllJournalForPersistenceId(persistenceId).filter(_.deleted === true).take(1).map(_.sequenceNumber).max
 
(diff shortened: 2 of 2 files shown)


lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
```diff
@@ -52,10 +52,10 @@ class JournalQueries(
       .update(true)
 
   private def _highestSequenceNrForPersistenceId(persistenceId: Rep[String]): Rep[Option[Long]] =
-    selectAllJournalForPersistenceId(persistenceId).map(_.sequenceNumber).max
+    selectAllJournalForPersistenceId(persistenceId).take(1).map(_.sequenceNumber).max
 
   private def _highestMarkedSequenceNrForPersistenceId(persistenceId: Rep[String]): Rep[Option[Long]] =
-    selectAllJournalForPersistenceId(persistenceId).filter(_.deleted === true).map(_.sequenceNumber).max
+    selectAllJournalForPersistenceId(persistenceId).filter(_.deleted === true).take(1).map(_.sequenceNumber).max
```
viktorklang, line 58: is `.max` needed if the result is already ordered by the id and we take the first?

core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
```diff
@@ -52,10 +52,10 @@ class JournalQueries(
       .update(true)
 
   private def _highestSequenceNrForPersistenceId(persistenceId: Rep[String]): Rep[Option[Long]] =
-    selectAllJournalForPersistenceId(persistenceId).map(_.sequenceNumber).max
+    selectAllJournalForPersistenceId(persistenceId).take(1).map(_.sequenceNumber).max
```
viktorklang, line 55: is `.max` needed if the result is already ordered by the id and we take the first?

core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
jaceksokol: It's not, frankly speaking, I have some troubles with it and I need to educate myself on how to do this in slick in this case.

octonato: What happens here is that this is a description of the query. 

The query returns `Query[Rep[Long], Long, Seq]`, in Slick parlance it means a query that will return `Seq[Long]`. 

Slick can't know we passed `1` to `take`, that's a runtime value from Slick point of view. It could have been `a` or `num`. Scala knows is hard code, but Slick doesn't. Anyway, moving on...

When we call `.max`, Slick knows that we can only return none or one value. It can then transform the query to a Rep, in that case `Rep[Long]`. 

If Slick had a `takeHead` or `takeFirst` we would be able to replace `take(1).max`, but that's the not case. 

Therefore we need the `.max`. 

Alternatively, we can drop it and change the method signature to ` Query[Rep[Long], Long, Seq]`, but then, when we call it we will have to do:

```scala
 db.run(queries.highestSequenceNrForPersistenceId(persistenceId).result.headOption)
```

I agree that's strange to use `max`, but it's better that this method is self-contained.

octonato (approved):
LGTM 

@jaceksokol, thanks for the contribution. 

Please check the CLA message. This is necessary to be able to merge this PR.

lightbend-cla-validator:
Hi @jaceksokol,

jaceksokol:
Looks like all the checks except one have stuck.

johanandren (approved):

PR #544, opened 2021-06-08
Initial version of JDBC implementation of DurableState

debasishg:
References https://github.com/akka/akka/issues/30277

Still a WIP but opening a PR for early review feedbacks. Contains the following : 

* JDBC implementation of `DurableStateStore` (javadsl and scaladsl)
* Tests including one with payload having a custom serializer
* Tests run against an H2 database
* JDBC implementation of `DurableStateStoreQuery` (javadsl and scaladsl)
* Tests for `currentChanges` and `changes` APIs against H2 and Postgres database
* `DurableStateSequenceActor` implemented for monitoring gaps in offsets
* Integration tests

To run tests:

```scala
// runs against h2 in memory database
$ sbt testOnly akka.persistence.jdbc.state.scaladsl.H2DurableStateSpec
...


// Integration test
// runs against Postgres database : 
// Ensure a Postgres instance running having a user named `docker` and a database named `docker`
$ It/testOnly akka.persistence.jdbc.integration.PostgresScalaJdbcDurableStateStoreQueryTest
```


core/src/main/scala/akka/persistence/jdbc/state/javadsl/DurableStateStore.scala
```diff
@@ -0,0 +1,55 @@
...
+          GetObjectResult(
+            Optional
+              .ofNullable(AkkaSerialization.fromRow(serialization)(row).toOption.getOrElse(null))
+              .asInstanceOf[Optional[A]],
+            row.seqNumber)
+
+        case None =>
+          throw new Exception(s"State object creation failed during fetch from store: persistenceId $persistenceId")
+      }
+    })
```
patriknw, line 39: Instead of implementing those twice for javadsl and scaladsl it should delegate to the scaladsl implementation and only do the conversions here. `scaladsl.DurableStateStore` can be passed as constructor parameter and those methods can delegate to that.

patriknw (commented):
looking good

patriknw:
>  In upsert we have tag as an input. Should we store tags in table as well

yes, that was what I mentioned in https://github.com/akka/akka-persistence-jdbc/pull/544#commitcomment-51863158, but you can do that as a second step if you like

patriknw:
> Also, just curious, can there be only 1 tag ?

I suggest that we only support one tag, see https://github.com/akka/akka/issues/30277#issuecomment-856468577

patriknw:
> But upsert takes a seqNr as input - shouldn't we use that input instead of adding 1 to the current value ? Also adding 1 to the current value means we need to fetch the current value. So do we need to have this fetch + update in a single transaction ?

Right, the seqNr parameter is the "new" seqNr. So it would looks something like this (pseudo):

```
if (seqNr == 1) {
  "insert... "
} else {
  val updatedRows = "update foo ... where seq_nr = ${seqNr - 1}" 
  if (updatedRows != 1)
    throw boom
}
```

debasishg:
@patriknw regarding the following ..

```scala 
if (seqNr == 1) {
  "insert... "
} else {
  val updatedRows = "update foo ... where seq_nr = ${seqNr - 1}" 
  if (updatedRows != 1)
    throw boom
}
```

The above code is not transactionally safe. Someone can insert a row between the check `if (seqNr ==1)` and doing the insert. Slick supports `insertAndUpdate` which will do transactional `upsert` if the underlying database supports it. I have used `insertAndUpdate` at the JDBC level. And at the top level I throw if `seqNr` is < 0. So first time insert and subsequent updates will be taken care of at the JDBC level if u supply a valid sequence number.

core/src/main/scala/akka/persistence/jdbc/state/javadsl/DurableStateStore.scala
debasishg: ok, didn't want to make `javadsl` depend on `scaladsl` package. But will change ..

debasishg: done ..

patriknw:
great if `insertAndUpdate` will handle it

debasishg:
@patriknw I added some unit tests as well. The ones added use `DurableStateStore[String]`. I will be working on some more with custom serializers. Please take a look when you have some time.

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/DurableStateStore.scala
```diff
@@ -0,0 +1,51 @@
...
+  val queries = new DurableStateQueries(profile, durableStateConfig)
+
+  def getObject(persistenceId: String): Future[GetObjectResult[A]] =
+    db.run(queries._selectByPersistenceId(persistenceId).result).map { rows =>
+      rows.headOption match {
+        case Some(row) =>
+          GetObjectResult(AkkaSerialization.fromRow(serialization)(row).toOption.asInstanceOf[Option[A]], row.seqNumber)
+
+        case None =>
+          throw new Exception(s"State object creation failed during fetch from store: persistenceId $persistenceId")
```
patriknw, line 30: This shouldn't throw. It should be None in the `GetObjectResult` value. In that case you can use 0 as seqNr.

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
```diff
@@ -0,0 +1,19 @@
...
+
+class DurableStateQueries(val profile: JdbcProfile, override val durableStateTableCfg: DurableStateTableConfiguration)
+    extends DurableStateTables {
+  import profile.api._
+
+  def _selectByPersistenceId(persistenceId: Rep[String]) =
+    durableStateTable.filter(_.persistenceId === persistenceId)
+
+  def _upsertDurableState(row: DurableStateTables.DurableStateRow) =
+    durableStateTable.insertOrUpdate(row)
```
patriknw, line 14: will this automatically verify the expected seqNr?

core/src/test/scala/akka/persistence/jdbc/state/JdbcDurableStateSpec.scala
```diff
@@ -0,0 +1,161 @@
...
+          _ = g.value shouldBe Some("a valid string")
+          u <- stateStoreString.upsertObject("p234", 2, "updated valid string", "t123")
+          _ = u shouldBe akka.Done
+          h <- stateStoreString.getObject("p234")
+
+        } yield h
+      } { v =>
+        v.value shouldBe Some("updated valid string")
+      }
+    }
```
patriknw, line 94: add another test when wrong seqNr is used (accidental concurrent writers)

patriknw (commented):
Looking good, a few comments. I'd like @octonato to review this to make sure that it conforms to existing conventions in this plugin.

core/src/main/scala/akka/persistence/jdbc/state/DurableStateTables.scala
```diff
@@ -0,0 +1,42 @@
...
+      stateSerId: Int,
+      stateSerManifest: Option[String])
+}
+
+trait DurableStateTables {
+  val profile: slick.jdbc.JdbcProfile
+  import profile.api._
+  def durableStateTableCfg: DurableStateTableConfiguration
+  // TODO:
+  // def tagTableCfg: DurableStateTagTableConfiguration
```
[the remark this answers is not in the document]
debasishg: ok .. will remove it and add tag as a column

core/src/main/scala/akka/persistence/jdbc/state/javadsl/DurableStateStore.scala
```diff
@@ -0,0 +1,32 @@
...
+import java.util.concurrent.CompletionStage
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.ExecutionContext
+import slick.jdbc.JdbcProfile
+import akka.Done
+import akka.persistence.state.javadsl.{ DurableStateUpdateStore, GetObjectResult }
+import akka.persistence.jdbc.state.DurableStateQueries
+import akka.persistence.jdbc.config.DurableStateTableConfiguration
+
+class DurableStateStore[A](
```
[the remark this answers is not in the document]
debasishg: ok, thought since we are already in the package `jdbc` we may not need it. Will add ..

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/DurableStateStore.scala
```diff
@@ -0,0 +1,52 @@
...
+import akka.persistence.jdbc.config.DurableStateTableConfiguration
+import akka.persistence.jdbc.state.DurableStateTables
+import akka.dispatch.ExecutionContexts
+
+class DurableStateStore[A](
+    db: JdbcBackend#Database,
+    profile: JdbcProfile,
+    durableStateConfig: DurableStateTableConfiguration,
+    serialization: Serialization)(implicit ec: ExecutionContext)
+    extends DurableStateUpdateStore[A] {
```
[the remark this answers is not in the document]
debasishg: I need to understand a bit more on what exactly does this class `DurableStateStoreProvider` offer ? We had one in in memory implementation as well IIRC

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/DurableStateStore.scala
debasishg: ok, will change ..

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
debasishg: What this does is if the record with the primary key `persistenceId` does not exist, it inserts. Otherwise updates. And it does this _transactionally_. Do we need any other validation ? Because if we use an if statement here the total operation will not be transactional. As per the spanner implementation ..

```scala
if (seqNr == 1) {
  "insert... "
} else {
  val updatedRows = "update foo ... where seq_nr = ${seqNr - 1}" 
  if (updatedRows != 1)
    throw boom
}
``` 

2 writers may try to execute the above code. Both pass the if check and tries to `insert`. Since the primary key is the `persistenceId` both will go through and we will have 2 records with `seqNr` = 1. Please suggest what logic should we follow here. I implemented the basic `upsert` support that Slick offers if the underlying database implements an upsert.

core/src/test/scala/akka/persistence/jdbc/state/JdbcDurableStateSpec.scala
debasishg: You mean 2 writers trying to write with the same sequence number and same persistence id ? What is the expected behavior here ? Note currently the primary key on the table is `persistence_id` ..

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/DurableStateStore.scala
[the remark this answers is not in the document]
patriknw: That is the factory of the javadsl/scaladsl DurableStateStore. It's the thing that is configured, see corresponding test in Akka https://github.com/akka/akka/blob/dev-durable-state/akka-persistence-typed-tests/src/test/scala/akka/persistence/typed/state/scaladsl/DurableStateBehaviorReplySpec.scala#L27

It can be very similar to https://github.com/akka/akka-persistence-jdbc/blob/master/core/src/main/scala/akka/persistence/jdbc/query/JdbcReadJournalProvider.scala

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
patriknw: I don't know what slick provides but in the end it would be good if the update could include the where condition on the expected seqNr. `"update foo ... where seq_nr = ${seqNr - 1}"`.

This kind of update condition is often called "optimistic" locking if you can search for if Slick has any built in support for that.

I understand the problem with seqNr 1, but if we can't have both checks I prefer to have it on the updates. It's a best effort check to detect concurrent writers, and if an application has that problem it's more likely that we detect it for updates (seqNr > 1) than for the first insert.

core/src/test/scala/akka/persistence/jdbc/state/JdbcDurableStateSpec.scala
patriknw: second write should fail if it doesn't have the expected sequence number

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
debasishg: @patriknw a few more questions ..

* For the update statement we need to check for the persistenceId match in the where clause as well - right ? e.g. `update state ... where persistence_id = $persistenceId and sequence_number = ${seqNr - 1}`

* Also if we make a composite primary key `persistence_id + sequence_number` then the possibility of inserting 2 records with the same sequence number is eliminated at the database level. So if the `insert` is executed concurrently for the same `persistence_id + sequence_number` one of them will fail - does this sound ok ?

* And in the update above what happens if the number of updated rows is `0` i.e. an incorrect sequence number is provided by the user - should we throw ? And note that there cannot be multiple records being updated as we cannot have multiple instances of the same `persistence_id + sequence_number` combination.

core/src/main/scala/akka/persistence/jdbc/state/DurableStateTables.scala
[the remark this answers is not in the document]
debasishg: @patriknw will `tag` be a mandatory (`NOT NULL` ) column or an optional (`NULL`) column ?

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
patriknw: > persistenceId match in the where clause as well

absolutely, that's the primary key

> composite primary key persistence_id + sequence_number

but then it will not be updates the existing row if it's a new primary key for each seqNr increment

> an incorrect sequence number is provided by the user - should we throw ?

yes, that should throw, look at Spanner implementation

> note that there cannot be multiple records being updated

right, the persistenceId is the primary key and there can be 0 or 1 rows for a given persistenceId

patriknw: Surprisingly there seems to be no support for optimistic locking in Slick https://github.com/slick/slick/issues/378

We could implement it with [plain sql](https://scala-slick.org/doc/3.3.3/sql.html) but let's skip this feature of checking the sequence number altogether. We want to move away from Slick anyway and then we can implement it with that library.

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/DurableStateStore.scala
```diff
@@ -0,0 +1,51 @@
...
+      AkkaSerialization.serialize(serialization, value).map { serialized =>
+        DurableStateTables.DurableStateRow(
+          persistenceId,
+          serialized.payload,
+          seqNr,
+          serialized.serId,
+          (if (serialized.serManifest.isEmpty()) None else Some(serialized.serManifest)))
+      }
+
+    Future.fromTry(row).map(queries._upsertDurableState).flatMap(db.run).map(_ => Done)(ExecutionContexts.parasitic)
```
octonato, line 46: we can avoid the map/flatMap (1 scheduling vs. 2)

eg:

```scala
Future.fromTry(row)
  .flatMap(r => db.run(queries._upsertDurableState(r)))
  .map(_ => Done)(ExecutionContexts.parasitic)
```

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
octonato: As far as I can tell, Slick does not provide any out-of-the-box optimistic locking, but we can implement if with a where clause in the update. 

> For the update statement we need to check for the persistenceId match in the where clause as well - right ? e.g. update state ... where persistence_id = $persistenceId and sequence_number = ${seqNr - 1}

Yes, in that case we need both. 

> Also if we make a composite primary key persistence_id + sequence_number then the possibility of inserting 2 records with the same sequence number is eliminated at the database level. So if the insert is executed concurrently for the same persistence_id + sequence_number one of them will fail - does this sound ok ?

I think this is ok, we let the DB take care of that. 

> And in the update above what happens if the number of updated rows is 0 i.e. an incorrect sequence number is provided by the user - should we throw ? And note that there cannot be multiple records being updated as we cannot have multiple instances of the same persistence_id + sequence_number combination.

The user in our case is the actor that is tracking the `seq_num`, not the real user. 
We can run into a situation where two actors are in memory, pointing to the same state and holding the same seq_num, for example on a badly configured cluster or a cluster that has a split brain. 

In that case, the optimistic locking will be the last change to avoid it. One of the actors will crash and that's good.

octonato (commented):
Looking good @debasishg. Thanks. 

I left a few comments and suggestions.

debasishg:
@patriknw, @octonato Made the following changes:

* added `tag` as a column in table
* reordered column names as per suggestion
* changed logic in `upsertObject` - got rid of Slick's `insertAndUpdate` and incorporated logic based on sequence number
* added `JdbcDurableStateStoreProvider` class
* some refactoring in slick code
* some class renames as per suggestion 
* added more tests as per suggestion

Please take another look ..

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
```diff
@@ -0,0 +1,72 @@
...
+        .flatMap(r => db.run(queries._insertDurableState(r)))
+        .map(_ => Done)(ExecutionContexts.parasitic)
+    } else {
+      // if seqNr > 1 we always try update and if that fails (returns 0 affected rows) we throw
+      Future
+        .fromTry(row)
+        .flatMap(r => db.run(queries._updateDurableState(r)))
+        .map { rowsAffected =>
+          if (rowsAffected == 0)
+            throw new Exception(
```
patriknw, line 63: I'd prefer `RuntimeException`

patriknw: or even `IllegalStateException`

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
```diff
@@ -0,0 +1,72 @@
...
+              s"Incorrect sequence number $seqNr provided: $seqNr has to be 1 more than the value existing in the database for this $persistenceId")
```
patriknw, line 64:
```suggestion
              s"Incorrect sequence number [$seqNr] provided: It has to be 1 more than the value existing in the database for persistenceId [$persistenceId]")
```

core/src/test/scala/akka/persistence/jdbc/state/JdbcDurableStateSpec.scala
```diff
@@ -0,0 +1,181 @@
...
+    "fail inserting incorrect sequence number" in {
+      whenReady {
+        (for {
+
+          n <- stateStoreString.upsertObject("p345", 1, "a valid string", "t123")
+          _ = n shouldBe akka.Done
+          g <- stateStoreString.getObject("p345")
+          _ = g.value shouldBe Some("a valid string")
+          u <- stateStoreString.upsertObject("p345", 1, "updated valid string", "t123")
```
patriknw, line 108: good, you can also add another test that tries to upsert seqNr 2 twice, which should fail

patriknw (commented):
great that the optimistic locking could be added

core/src/test/scala/akka/persistence/jdbc/state/JdbcDurableStateSpec.scala
debasishg: done ..

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
debasishg: changed

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
debasishg: done

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
```diff
@@ -0,0 +1,31 @@
...
+  def _updateDurableState(row: DurableStateTables.DurableStateRow) = {
+    durableStateTable
+      .filter(r =>
+        r.persistenceId === row.persistenceId &&
+        r.seqNumber === row.seqNumber - 1)
+      .map(r => (r.statePayload, r.stateSerId, r.stateSerManifest))
+      .update((row.statePayload, row.stateSerId, row.stateSerManifest))
+  }
+
+  def _upsertDurableState(row: DurableStateTables.DurableStateRow) =
```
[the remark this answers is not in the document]
debasishg: done ..

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
```diff
@@ -0,0 +1,31 @@
...
+
+class DurableStateQueries(val profile: JdbcProfile, override val durableStateTableCfg: DurableStateTableConfiguration)
+    extends DurableStateTables {
+  import profile.api._
+
+  def _selectByPersistenceId(persistenceId: Rep[String]) =
+    durableStateTable.filter(_.persistenceId === persistenceId)
+
+  def _insertDurableState(row: DurableStateTables.DurableStateRow) =
+    durableStateTable += row
```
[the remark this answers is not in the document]
debasishg: 👍

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  def _updateDurableState(row: DurableStateTables.DurableStateRow) = {
+    durableStateTable
+      .filter(r =>
+        r.persistenceId === row.persistenceId &&
+        r.seqNumber === row.seqNumber - 1)
```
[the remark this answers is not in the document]
debasishg: yes!

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
```diff
@@ -60,8 +60,8 @@ class JdbcDurableStateStore[A](
         .flatMap(r => db.run(queries._updateDurableState(r)))
         .map { rowsAffected =>
           if (rowsAffected == 0)
-            throw new Exception(
-              s"Incorrect sequence number $seqNr provided: $seqNr has to be 1 more than the value existing in the database for this $persistenceId")
+            throw new IllegalArgumentException(
+              s"Incorrect sequence number [$seqNr] provided: It has to be 1 more than the value existing in the database for persistenceId $persistenceId")
```
octonato: I think this message is targeting the final user, but the final user is not really in control of which seqNr is being passed. So they can't control that it needs to be 1 more than the value in the DB.

We will run into this situation if we have a cluster split brain or a user using the `DurableStateBehavior` without a cluster and not taking care of duplicated instances. 

Instead, we can be more explicit about it, something like: 

> Incorrect sequence number [$seqNr] provided. Another instance of that same durable state actor is probably running in your system.

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  def upsertObject(persistenceId: String, seqNr: Long, value: A, tag: String): Future[Done] = {
+    require(seqNr > 0)
+    val row =
+      AkkaSerialization.serialize(serialization, value).map { serialized =>
+        DurableStateTables.DurableStateRow(
+          persistenceId,
+          seqNr,
+          serialized.payload,
+          if (tag.isEmpty()) None else Some(tag),
```
octonato, line 42: Are we sure `tag` won't be null? 

Maybe 

```suggestion
          Option(tag).filter(_.trim.nonEmpty),
```

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
```diff
@@ -0,0 +1,28 @@
...
+      .map(r => (r.statePayload, r.stateSerId, r.stateSerManifest))
+      .update((row.statePayload, row.stateSerId, row.stateSerManifest))
```
octonato, line 22: The `upsertObject` method accepts a `tag`, but we are not using it in the update. 

We don't recommend updating tags for event sourced entities, because that breaks the projections. The right thing to do for ES is to re-tag the whole journal and restart the consumers.

However, for durable state we may uncover an interesting use case. 

We can start tagging the `User` entity with `user0`, `user1` and `user2` (using persistenceId.hashCode % 3). And we have three projections consuming those tags. 

We can then change the code to use `% 10` and we start tagging from `user0` to `user9`. 

In the beginning, we will have 10 projections running but only 3 doing real work. As we start to update the `Users`, the start to migrate to another tag slice. After some time, all 10 projections will be doing work.

octonato (commented):
Happy to see we got the optimistic locking in a quite straightforward way. That will be quite cool to have durable state actors. 


@debasishg, have a look on how we deal with integration tests, for example here:
https://github.com/akka/akka-persistence-jdbc/blob/master/core/src/it/scala/akka/persistence/jdbc/integration/AllPersistenceIdsTest.scala

Each time we have the `h2` one that is run in `test` and then the other DBs are in `it:test` and they run with docker. 

In the scripts folder, we have some launch scripts to start the dockers for each DB individually and then you can run `sbt It/testOnly akka.persistence.jdbc.integration.Postgres*"` to pick the DB you want to test. 

It will be useful to add the integration tests as soon as possible and verify if the update is behaving the way we want, ie: returning 0 affected rows when 'failing'.

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
debasishg: Suggest adding both the messages like 
```scala
throw new IllegalArgumentException(              
  s"Incorrect sequence number [$seqNr] provided: It has to be 1 more than the value existing in the database for persistenceId $persistenceId. Maybe another instance of that same durable state actor is running in your system")
```

@patriknw @octonato ?

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
debasishg: done

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
debasishg: added tag in the list of columns to be updated ..

debasishg: also I guess the `stateTimestamp` column needs to be updated on upsert - @patriknw ?

debasishg:
@patriknw @octonato Just pushed in an initial implementation of `JdbcDurableStateStoreQuery`. Could you please take  a look. I will soon start writing the tests but it would be good to have another pair of eyes at the implementation before that.

core/src/test/scala/akka/persistence/jdbc/state/JdbcDurableStateSpec.scala
debasishg: added ..

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStoreQuery.scala
```diff
@@ -0,0 +1,79 @@
...
+import akka.stream.scaladsl.{ Sink, Source }
+import slick.jdbc.{ JdbcBackend, JdbcProfile }
+
+class JdbcDurableStateStoreQuery[A](
+    stateStore: JdbcDurableStateStore[A],
+    db: JdbcBackend#Database,
+    profile: JdbcProfile,
+    durableStateConfig: DurableStateTableConfiguration,
+    serialization: Serialization)(implicit m: Materializer, e: ExecutionContext)
+    extends DurableStateStoreQuery[A] {
```
patriknw, line 24: I think we can have one implementation that `extends DurableStateUpdateStore with DurableStateStoreQuery`, i.e. these addition can be added to the `JdbcDurableStateStore` class instead.

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStoreQuery.scala
```diff
@@ -0,0 +1,79 @@
...
+              }
+            }.max
+            /* if retrieved set is empty, then wait else continue */
+            val nextControl = if (chgs.isEmpty) ContinueDelayed else Continue
+            Some(((Offset.sequence(nextStartOffset), nextControl), chgs))
+          }
+        }
+        retrieveNextBatch()
+      }
+      .mapConcat(identity)
```
patriknw, line 57: I assume this follows the same pattern as existing eventsByTag implementation.

debasishg: Yes. Just wanted to confirm one thing. In the `JdbcReadJournal#eventsByTag` implementation, it also uses an actor to find out the `MaxOrderId` till which query will be done. And once that max is reached, it ends the stream. Since you mentioned that this is going to be a live stream I have excluded that part.

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStoreQuery.scala
debasishg: done ..

debasishg:
@patriknw Done the changes that you suggested. Please take  a look when u have some time. The `JdbcDurableStateStore` now has the query implementation as well. And I have implemented the `changes` function along the same lines as the one in `eventsByTag` in `JdbcReadJournal` subject to some caveats that I mentioned in the last comment.

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStoreQuery.scala
```diff
@@ -30,7 +33,29 @@ class JdbcDurableStateStoreQuery[A](
...
+  }
```
johanandren: I don't see that `control` is used anywhere, how does that `ContinueDelayed` delaying work, kind of looks like it will now just hit the `currentChanges` as fast as it can if there is downstream demand but no new results?

debasishg: Hi @johanandren - Possibly you are looking at an outdated commit .. I see this in the branch .. https://github.com/akka/akka-persistence-jdbc/blob/wip-durable-state/core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala#L114-L119

johanandren: Ah, that explains it, sorry for the noise, got tricked by github!

debasishg:
@patriknw @octonato - Could you please take another look at the implementation ? I have tests based on h2 that run ok. I will be working on the Postgres based implementation from tomorrow. (I am on well-being day leave today)

core/src/main/scala/akka/persistence/jdbc/config/AkkaPersistenceConfig.scala
```diff
@@ -210,7 +210,9 @@ class DurableStateTableColumnNames(config: Config) {
 class DurableStateTableConfiguration(config: Config) {
   private val cfg = config.getConfig("tables.state")
   val refreshInterval: FiniteDuration = config.asFiniteDuration("refreshInterval")
+  val batchSize: Int = config.getInt("batchSize")
   val stopAfterEmptyFetchIterations: Int = config.getInt("stopAfterEmptyFetchIterations")
+  val schemaType: String = config.getString("schemaType")
```
johanandren, line 215: Maybe worth failing fast on unsupported rather than later with a match error?

core/src/main/scala/akka/persistence/jdbc/state/SequenceNextValUpdater.scala
```diff
@@ -0,0 +1,41 @@
+package akka.persistence.jdbc.state
+
+import slick.jdbc.JdbcProfile
+import slick.dbio.Effect
+import slick.sql.SqlStreamingAction
+
+trait SequenceNextValUpdater {
+  def nextValUpdater(sequenceName: String): String
+  def _getSequenceName(): SqlStreamingAction[Vector[String], String, Effect]
```
johanandren, line 9: No underscore in the name since its not private but actual SPI? I guess the entire trait is internal `private[jdbc] ` anyway?

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/package.scala
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (C) 2014 - 2019 Dennis Vriend <https://github.com/dnvriend>
+ * Copyright (C) 2019 - 2021 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.persistence.jdbc.state
+
+import akka.persistence.query._
+package object scaladsl {
```
johanandren, line 9: Package objects going away for Scala 3 so better to not introduce new ones.

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
```diff
@@ -86,56 +87,90 @@ class JdbcDurableStateStore[A](
...
+  private def _currentChangesByTag(
+      tag: String,
+      from: Long,
+      batchSize: Long,
+      queryUntil: MaxOffset): Source[DurableStateChange[A], NotUsed] = {
+    if (queryUntil.value < from.value) Source.empty
+    else changesByTag(tag, from, queryUntil.value, batchSize).mapAsync(1)(Future.fromTry)
+  }
+
+  private def changesByTag(
```
johanandren, line 113: A bit confusing with one private method called `changesByTag` and one `_changesByTag`, giving them names that indicates how they are different would be nice.

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
```diff
@@ -86,56 +87,90 @@ class JdbcDurableStateStore[A](
...
+      .futureSource(maxJournalSequence().map { maxOrderingInDb =>
+        _changesByTag(tag, offset, terminateAfterOffset = Some(maxOrderingInDb))
+      })
+      .mapMaterializedValue(_ => NotUsed)
+  }
+
+  private def _currentChangesByTag(
+      tag: String,
+      from: Long,
+      batchSize: Long,
```
johanandren, line 107: `batchSize` here but in `changesByTag` it's rather the max number of results to collect?

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
```diff
@@ -86,56 +87,90 @@ class JdbcDurableStateStore[A](
...
+  }
 
-    // a collection to store the start offsets for every iteration
-    val lastStartOffsets = collection.mutable.ListBuffer.empty[Long]
+  private def _changesByTag(
+      tag: String,
+      offset: Long,
+      terminateAfterOffset: Option[Long]): Source[DurableStateChange[A], NotUsed] = {
+    val batchSize = durableStateConfig.batchSize
+    val startingOffsets = collection.mutable.ListBuffer.empty[Long]
```
johanandren, line 138: This is used in a future callback (on other threads) so it is not safe to share a mutable structure like this. Would need to be an immutable collection passed as unfold state or a thread safe collection of some sort.

johanandren (commented):
Some nitpicks and one concurrency issue.

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  private[jdbc] def _getSequenceName() = sequenceNextValUpdater._getSequenceName()
+
+  def _delete(persistenceId: String) = {
+    durableStateTable.filter(_.persistenceId === persistenceId).delete
+  }
+
+  private[jdbc] val maxOffsetQuery = Compiled {
+    durableStateTable.map(_.globalOffset).max.getOrElse(0L)
```
patriknw, line 57: will this be changed to use the same approach as in the journal, using such `journalSequenceActor`?

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
```diff
@@ -0,0 +1,209 @@
...
+
+                // We may also stop if we have found an event with an offset >= target
+                case Some(target) if xs.exists(_.offset.value >= target) => Stop
+
+                // otherwise, disregarding if Some or None, we must decide how to continue
+                case _ =>
+                  if (
+                    stopAfterEmptyFetchIterations > 0 && equalLastNElements(
+                      startingOffsets.toList,
+                      stopAfterEmptyFetchIterations)
```
patriknw, line 161: this case is only intended for tests, wonder if those tests could be written in another way to avoid this additional stuff and the mutable `startingOffsets`

core/src/main/scala/akka/persistence/jdbc/state/SequenceNextValUpdater.scala
debasishg: done .. the trait is also marked `private[jdbc]`

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/package.scala
debasishg: done ..

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
debasishg: changed the names ..

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
debasishg: `max` changed to `batchSize`

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
debasishg: @patriknw The problem is the `source` in this case is live and hence when I do a `source.runWith(Sink.ignore)` I get a `Future` which does not complete. The API `changes(persistenceId, tag)` does not give me any handle to complete the future. And also future cancelation is a hairy task. Hence for tests, I had to resort to such tactics for completing the future. Any suggestions welcome ..

core/src/main/scala/akka/persistence/jdbc/config/AkkaPersistenceConfig.scala
johanandren: I'd be fine with stringly typed here, but if it is not a lot of work actual type here would be better ofc.

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
johanandren: But then it is used as max in the actual query (limiting the results rather than deciding how large each batch is)?

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
johanandren: You could perhaps use `takeWhile` or `take` to limit downstream in the test instead?

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
debasishg: @johanandren The `max` is used to indicate the maximum number of rows that the query will fetch. That's equal to the `batchSize` in our case. Hence keeping the name uniform as `batchSize` looks ok I thought.

core/src/main/scala/akka/persistence/jdbc/config/AkkaPersistenceConfig.scala
debasishg: @johanandren Just one thought .. if I do the check in the config file, even then I will have to repeat that check in `DurableStateQueries` where I instantiate the proper type of `SequenceValUpdater` ..

```scala
val sequenceNextValUpdater = durableStateTableCfg.schemaType match {
  case "H2"       => new H2SequenceNextValUpdater(profile)
  case "POSTGRES" => new PostgresSequenceNextValUpdater(profile)
}
```

But still we can fail fast at the config level at the expense of a double check ..

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
debasishg: @johanandren Even if I make `takeWhile` or `take` where would I do that ? All queries are in the main code and not in tests. The only API I have in tests in `changes(persistencId, tag)`. Hence my logic was if I get a preset number of 0 results in query I stop.

core/src/main/scala/akka/persistence/jdbc/config/AkkaPersistenceConfig.scala
johanandren: Yeah, I think double check is fine, was thinking mostly about the user experience when accidentally having wrong config and that it would give a good error saying "A and B is supported but not C which you specified"

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
johanandren: Ah, ok, thanks for clarifying!

core/src/main/scala/akka/persistence/jdbc/config/AkkaPersistenceConfig.scala
debasishg: done ..

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
debasishg: Yes, I will change this to use `JournalSequenceActor` ..

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
debasishg: Now changed .. immutability all the way ..

debasishg:
@patriknw @johanandren @octonato May be another look ? I have the actor in place now to monitor the monotonicity. Addressed other issues raised in the PR feedback.

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/DurableStateSequenceActor.scala
```diff
@@ -0,0 +1,186 @@
...
+    def empty: MissingElements = MissingElements(Vector.empty)
+  }
+}
+
+/**
+ * To support the changesByTag query, this actor keeps track of which rows are visible in the database.
+ * This is required to guarantee the changesByTag does not skip any rows in case rows with a higher (ordering) id are
+ * visible in the database before rows with a lower (ordering) id.
+ */
+class DurableStateSequenceActor[A](stateStore: JdbcDurableStateStore[A], config: DurableStateSequenceRetrievalConfig)(
```
octonato, line 57: I don't think this will do the trick for the durable state. The `JournalSequenceActor` was designed to detect gaps that occur when a more recent ordering (`n`) shows up on the table before  a previous one (`n-1`). 

Although that could also happen for durable state, we should have something that can also detect missing updates. 

If we have:

```scala
1 | abc
2 | def
```

And both get updated, `abc` becomes 3 and `def` becomes 4. It's possible `def` transaction commits first and the query see the table as: 

```scala
1 | abc
4 | def
```

The stream delives 4, user saves the offset and now 3 will never be delivered. We lost an update.

octonato (commented):
@debasishg, I finally got some time to do a more detailed review. 

I think we will need to re-think how to deal with missing updates as they can still happen, it seems (see my comments).

core/src/main/resources/schema/h2/h2-create-schema.sql
```diff
@@ -39,3 +39,17 @@ CREATE TABLE IF NOT EXISTS "snapshot" (
     PRIMARY KEY("persistence_id","sequence_number")
     );
 
+CREATE TABLE IF NOT EXISTS "state" (
+    "global_offset" BIGINT NOT NULL AUTO_INCREMENT,
+    "persistence_id" VARCHAR(255) NOT NULL,
+    "sequence_number" BIGINT NOT NULL,
```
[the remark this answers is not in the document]
debasishg: @octonato Calling it `version` or `revision` sounds good to me. But since @patriknw used this term in the API definitions, I decided to use `sequence_number`. Maybe we should ping @patriknw once for this change ?

core/src/main/resources/schema/postgres/postgres-create-schema.sql
```diff
@@ -47,3 +47,16 @@ CREATE TABLE IF NOT EXISTS public.snapshot (
...
+    "sequence_number" BIGINT NOT NULL,
+    "state_payload" BYTEA NOT NULL,
+    "state_serial_id" INTEGER NOT NULL,
+    "state_serial_manifest" VARCHAR(255),
+    "tag" VARCHAR,
+    "state_timestamp" BIGINT NOT NULL,
+    PRIMARY KEY("persistence_id")
+    );
+CREATE INDEX CONCURRENTLY "state_tag_idx" on "state" ("tag");
+CREATE INDEX CONCURRENTLY "state_global_offset_idx" on "state" ("global_offset");
```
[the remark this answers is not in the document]
debasishg: @octonato I found it to be a recommended practice in Postgres. https://www.2ndquadrant.com/en/blog/create-index-concurrently/

> A unique distinguishing factor of CIC is that it can build a new index on the table, without blocking it from updates/inserts/deletes.

core/src/main/scala/akka/persistence/jdbc/testkit/internal/SchemaUtilsImpl.scala
```diff
@@ -37,8 +37,9 @@ private[jdbc] object SchemaUtilsImpl {
    * INTERNAL API
    */
   @InternalApi
-  private[jdbc] def dropIfExists(logger: Logger)(implicit actorSystem: ClassicActorSystemProvider): Future[Done] = {
-    val slickDb: SlickDatabase = loadSlickDatabase("jdbc-journal")
+  private[jdbc] def dropIfExists(logger: Logger, configKey: String = "jdbc-journal")(
```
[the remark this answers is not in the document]
debasishg: Initially thought I would use the same functions for a different config-key. Hence made it a parameter. But later did not use it. Now changed back to original.

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/DurableStateSequenceActor.scala
debasishg: @octonato let me first clear the confusion in terminology. What u r saying as `ordering` for the `JournalSequenceActor` it is the `global_offset` in `DurableStateSequenceActor` - am I correct ? So in the implementation of the actor, `OrderingId` refers to `ordering` for the journal and the `offset` for the state implementation. `JournalSequenceActor` detects missing `ordering`, while `DurableStateSequenceActor` detects missing `offsets`. Then why should we have a lost update in this implementation ?

debasishg: @octonato In other words why should the actor not detect that 3 is missing and deliver it as well ?

core/src/test/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateSpec.scala
```diff
@@ -0,0 +1,452 @@
...
+      chgs.size shouldBe 1
+      chgs.map(_.offset.value).max shouldBe 4
+
+      // upsert more and fetch from after the last offset
+      upsertManyFor(stateStoreString, "p1", "t1", 5, 7)
+      val moreChgs = currentChanges("t1", chgs.head.offset).runWith(Sink.seq).futureValue
+      moreChgs.size shouldBe 1
+      moreChgs.map(_.offset.value).max shouldBe 11
+
+      // upsert same tag, different peristence id and fetch from after the last offset
```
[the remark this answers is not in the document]
debasishg: done ..

core/src/test/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateSpec.scala
```diff
@@ -0,0 +1,452 @@
...
+
+class H2DurableStateSpec extends JdbcDurableStateSpec(ConfigFactory.load("h2-application.conf"), H2) {
+  implicit lazy val system: ActorSystem =
+    ActorSystem("JdbcDurableStateSpec", config.withFallback(customSerializers))
+}
+
+class PostgresDurableStateSpec extends JdbcDurableStateSpec(ConfigFactory.load("postgres-application.conf"), Postgres) {
+  implicit lazy val system: ActorSystem =
+    ActorSystem("JdbcDurableStateSpec", config.withFallback(customSerializers))
+}
```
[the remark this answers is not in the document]
debasishg: @octonato I have https://github.com/akka/akka-persistence-jdbc/blob/wip-durable-state/core/src/it/scala/akka/persistence/jdbc/integration/PostgresScalaJdbcDurableStateChangesByTagTest.scala in the integration test folder. Will remove this.

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
patriknw: yes, the timestamp should be updated

core/src/main/resources/schema/h2/h2-create-schema.sql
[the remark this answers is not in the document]
patriknw: We can rename it to `revision` and then it should be done in all places Akka DurableStateBehavior and store apis.

> optimistic locking implementation

It can be used for other things also, such as de-duplication in consumers of changes, but that works fine with naming it `revision`.

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/DurableStateSequenceActor.scala
patriknw: > JournalSequenceActor detects missing ordering, while DurableStateSequenceActor detects missing offsets

That is right.

I assume that Renato's example is using global_offset as the first column and persistence_id as the second column.

For that example, the DurableStateSequenceActor will detect that offset 3 is missing when it sees 4, it will query again, and will eventually find 3. 4 is not delivered until 3 has been seen.

However, there is a another scenario where we have a problem. If there are several quick updates for the same persistence_id we will see the intermediate updates as missing offsets. Those will be retried and give up on those after timeout. All good in the end, but it will drastically slow down the queries.

```
global_offset | persistence_id | revision
---------------------------------------
17            | abc            | 1
```
2 quick updates of "abc" and the query will see
```
global_offset | persistence_id | revision
---------------------------------------
19            | abc            | 3
```
missing offset 18

could be more complex with several persistence_ids updated quickly
```
global_offset | persistence_id | revision
---------------------------------------
17            | abc            | 1
18            | def            | 1
```
2 quick updates of abc and 2 of def
```
global_offset | persistence_id | revision
---------------------------------------
20            | abc            | 3
22            | def            | 3
```
missing offset 19 and 21

One way to solve that could be to also track seen persistence_id/revision. In above example, we could then see that a total of 2 revisions are missing and that matches 2 missing offsets. Then it should be safe to say that we have seen all changes.

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
debasishg: @johanandren Now fixed. Using `takeWhile` .. Thanks.

core/src/main/resources/reference.conf
```diff
@@ -576,6 +576,24 @@ jdbc-durable-state-store {
...
+    # In case a number in the sequence is missing, this is the ammount of retries that will be done to see
+    # if the number is still found. Note that the time after which a number in the sequence is assumed missing is
+    # equal to maxTries * queryDelay
+    # (maxTries may not be zero)
+    max-tries = 10
+    # How often the actor will query for new data
+    query-delay = 1 second
+    # The maximum backoff time before trying to query again in case of database failures
+    max-backoff-query-delay = 1 minute
+    # The ask timeout to use when querying the durable-state sequence actor, the actor should normally repond very quickly,
```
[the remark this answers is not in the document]
debasishg: done ..

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
```diff
@@ -112,52 +120,58 @@ class JdbcDurableStateStore[A](
...
-                  else ContinueDelayed
+      .unfoldAsync[(Long, FlowControl, List[Long]), Seq[DurableStateChange[A]]]((offset, Continue, startingOffsets)) {
+        case (from, control, s) =>
+          // println(s"start offset $from $s")
+          def retrieveNextBatch() = {
+            for {
+              queryUntil <- stateSequenceActor.ask(GetMaxOrderingId).mapTo[MaxOrderingId]
+              xs <- currentChangesByTag(tag, from, batchSize, queryUntil).runWith(Sink.seq)
+            } yield {
+              // println(s"queryuntil : ${queryUntil.maxOrdering} xs size : ${xs.size} pids : ${xs.map(_.persistenceId)}")
```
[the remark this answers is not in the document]
debasishg: 👍

core/src/test/scala/akka/persistence/jdbc/state/scaladsl/DataGenerationHelper.scala
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  def upsertParallel(store: JdbcDurableStateStore[String], pids: Set[String], tag: String, noOfItems: Int)(
+      implicit ec: ExecutionContext) = {
+
+    for {
+      _ <- Future.unit
+      f1 = Future(upsertManyFor(store, pids.head, tag, 1, noOfItems))(ExecutionContexts.parasitic)
+      f2 = Future(upsertManyFor(store, pids.tail.head, tag, 1, noOfItems))(ExecutionContexts.parasitic)
+      f3 = Future(upsertManyFor(store, pids.last, tag, 1, noOfItems))(ExecutionContexts.parasitic)
```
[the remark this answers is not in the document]
debasishg: Now fixed. Removed `parasitic` and using the `implicit ec`

core/src/main/resources/reference.conf
```diff
@@ -541,3 +541,131 @@ jdbc-read-journal {
...
+  class = "akka.persistence.jdbc.state.scaladsl.DurableStateStore"
+
+  # number of records fetched from the store at once
+  batchSize = 500
+  # New states are retrieved (polled) with this interval.
+  refreshInterval = "1s"
+  # stop the live stream query for changes after this number of iterations
+  # to be used only for tests
+  # a value of 0 means don't stop
+  stopAfterEmptyFetchIterations = 0
```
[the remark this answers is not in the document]
debasishg: Now removed ..

core/src/test/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateSpec.scala
```diff
@@ -0,0 +1,452 @@
...
+      implicit system =>
+        val stateStoreString =
+          new JdbcDurableStateStore[String](db, schemaTypeToProfile(schemaType), durableStateConfig, serialization)
+
+        import stateStoreString._
+
+        upsertParallel(stateStoreString, Set("p1", "p2", "p3"), "t1", 5)(e).futureValue
+        val source = changes("t2", NoOffset)
+        val m = collection.mutable.ListBuffer.empty[(String, Offset)]
+        val z = source.map(ds => m += ((ds.persistenceId, ds.offset))).runWith(Sink.ignore)
```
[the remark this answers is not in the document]
debasishg: Yes, the `takeWhile` trick worked :-)

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
debasishg: yes, doing that ..

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/DurableStateSequenceActor.scala
debasishg: @patriknw perhaps I am missing something. When we query for the `changes` we want to get the latest offset for a `persistence_id` for some specified `tag` and _NOT_ all the intermediate ones. Then what is the harm in the above case if we do not see offset 19 and 21 ?

patriknw: We can't know if 19 and 21 were written by abc and def respectively. They could have been written by xyz and then we can't ignore them.

debasishg: @patriknw But if they are of a different persistence_id, then the actor will eventually fetch them in the query. Maybe I am missing how can the actor miss them .. If 19 and 21 are from diff persistence_ids then they will be inserted and the query will get them. If they are from the same persistence id, then also the actor will get at least the max of them.

patriknw: That is right, with current implementation, but my point was:

> Those will be retried and give up on those after timeout. All good in the end, but it will drastically slow down the queries.

We will always see such timeout delays when there are quick updates to the same persistence_id. Since that seems like a rather common scenario it would be good to find a solution where we can avoid these delays (in most cases).

debasishg: ah .. ok .. makes sense now.

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  implicit val uuidSetter = SetParameter[Array[Byte]] { case (bytes, params) =>
+    params.setBytes(bytes)
+  }
+
+  private[jdbc] def selectFromDbByPersistenceId(persistenceId: Rep[String]) =
+    durableStateTable.filter(_.persistenceId === persistenceId)
+
+  private[jdbc] def insertDbWithDurableState(row: DurableStateTables.DurableStateRow, seqName: String) = {
```
patriknw, line 27: rename s/seqName/seqNextValue/

core/src/main/scala/akka/persistence/jdbc/state/DurableStateStoreProvider.scala
```diff
@@ -0,0 +1,37 @@
...
+    db: JdbcBackend#Database,
+    profile: JdbcProfile,
+    durableStateConfig: DurableStateTableConfiguration,
+    serialization: Serialization)(implicit val system: ExtendedActorSystem)
+    extends DurableStateStoreProvider {
+
+  implicit val ec: ExecutionContext = system.dispatcher
+  implicit val mat: Materializer = SystemMaterializer(system).materializer
+
+  override def scaladslDurableStateStore(): DurableStateStore[Any] =
```
patriknw, line 29: `val` instead of `def`

core/src/main/scala/akka/persistence/jdbc/state/DurableStateStoreProvider.scala
```diff
@@ -0,0 +1,37 @@
...
+    new scaladsl.JdbcDurableStateStore[Any](db, profile, durableStateConfig, serialization)
+
+  override def javadslDurableStateStore(): JDurableStateStore[AnyRef] =
```
patriknw, line 32: `val`

patriknw (commented):
just a few minor comments

core/src/main/resources/reference.conf
```diff
@@ -541,3 +541,126 @@ jdbc-read-journal {
...
+jdbc-durable-state-store {
+  class = "akka.persistence.jdbc.state.scaladsl.DurableStateStore"
+
+  # number of records fetched from the store at once
+  batchSize = 500
+  # New states are retrieved (polled) with this interval.
+  refreshInterval = "1s"
+  # can be H2, POSTGRES
+  # schemaType = H2
+  schemaType = Postgres
```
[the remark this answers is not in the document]
octonato: Is that so that we will only support it for Postgres and H2? Meaning, only Postgres (for prod)? 

> should it have a default value? does the journal have a default value for this?

I don't think so. Specially if we don't want/plan to support other DBs.

core/src/main/scala/akka/persistence/jdbc/state/AkkaSerialization.scala
```diff
@@ -0,0 +1,37 @@
...
+ * Copyright (C) 2014 - 2019 Dennis Vriend <https://github.com/dnvriend>
+ * Copyright (C) 2019 - 2021 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.persistence.jdbc.state
+
+/*
+ * Copyright (C) 2014 - 2019 Dennis Vriend <https://github.com/dnvriend>
+ * Copyright (C) 2019 - 2021 Lightbend Inc. <https://www.lightbend.com>
+ */
```
octonato, line 11: something went wrong here.

core/src/main/scala/akka/persistence/jdbc/state/AkkaSerialization.scala
```diff
@@ -0,0 +1,37 @@
...
+    val p2 = payload.asInstanceOf[AnyRef]
+    val serializer = serialization.findSerializerFor(p2)
+    val serManifest = Serializers.manifestFor(serializer, p2)
+    val serialized = serialization.serialize(p2)
+    serialized.map(payload => AkkaSerialized(serializer.identifier, serManifest, payload))
+  }
+
+  def fromRow(serialization: Serialization)(row: DurableStateTables.DurableStateRow): Try[AnyRef] = {
+    serialization.deserialize(row.statePayload, row.stateSerId, row.stateSerManifest.getOrElse(""))
+  }
```
octonato, line 36: Maybe better to have only one `AkkaSerialization` object. And we can have two `fromRow` methods. Either removing the currying or having explicit named for the `fromRow` method.

This is internal API, so we don't need to be cautious when making bin-incompatible changes.

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
```diff
@@ -0,0 +1,100 @@
...
+import akka.persistence.jdbc.config.DurableStateTableConfiguration
+
+class DurableStateQueries(val profile: JdbcProfile, override val durableStateTableCfg: DurableStateTableConfiguration)
+    extends DurableStateTables {
+  import profile.api._
+
+  lazy val sequenceNextValUpdater = durableStateTableCfg.schemaType match {
+    case "H2"       => new H2SequenceNextValUpdater(profile)
+    case "Postgres" => new PostgresSequenceNextValUpdater(profile)
+  }
```
octonato, line 18: Are we planning to add the others or have we discovered that the trick we do with the seqNum is not possible with the other DBs? 

Otherwise, we can add a catch-all and throw an informative exception.

octonato (commented):
@debasishg, I left a few more comments and suggestions. 

Thanks for all that work and patience. 😄

core/src/main/resources/reference.conf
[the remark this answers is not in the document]
debasishg: @patriknw @octonato The `schemaType` needs to have a value since we have some platform specific code in the implementation. The statement that fetches the next value of the sequence differs between Postgres and H2 (haven't yet looked at other implementations though). Hence when building the sequence updater I have code like this ..

```scala
lazy val sequenceNextValUpdater = durableStateTableCfg.schemaType match {
  case "H2"       => new H2SequenceNextValUpdater(profile)
  case "Postgres" => new PostgresSequenceNextValUpdater(profile)
}
```

Hence the config needs to have the `schemaType` with a default value. This can of course be changed by changing the config when required (e.g. to run tests).

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
```diff
@@ -0,0 +1,212 @@
...
+            case Stop     => Future.successful(None)
+            case Continue => retrieveNextBatch()
+            case ContinueDelayed =>
+              akka.pattern.after(durableStateConfig.refreshInterval, system.scheduler)(retrieveNextBatch())
+          }
+      }
+      .mapConcat(identity)
+  }
+
+  private[jdbc] def maxStateStoreSequence(): Future[Long] =
```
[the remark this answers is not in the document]
debasishg: done ..

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
debasishg: done ..

core/src/main/scala/akka/persistence/jdbc/state/DurableStateStoreProvider.scala
debasishg: done ..

core/src/main/scala/akka/persistence/jdbc/state/DurableStateStoreProvider.scala
debasishg: done ..

core/src/main/scala/akka/persistence/jdbc/state/AkkaSerialization.scala
debasishg: fixed ..

core/src/main/scala/akka/persistence/jdbc/state/AkkaSerialization.scala
debasishg: Refactored to have 1 `AkkaSerialization` object 

* Moved `AkkaSerialization` to `akka.persistence.jdbc`. It was in `akka.persistence.jdbc.journal.dao`. Didn't want to have `state` depend on `journal`. Hence the move
* Added a method `fromDurableStateRow` to the object `AkkaSerialization` to be used from `state`. Changed the name to avoid any ambiguity.
* Removed the other `AkkaSerialization` object

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
debasishg: We will add other dbs as well. I am sure the scheme will work for other dbs. But since we decided to test for H2 and Postgres for the time being, I kept it like that. Added a `case_ => ???` that will throw for now.

debasishg:
@octonato @patriknw Incorporated all the suggestions raised so far. Another look ?

core/src/main/resources/reference.conf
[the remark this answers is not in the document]
patriknw: The user should only choose database at one place, which is the Slick profile: `akka-persistence-jdbc.shared-databases.slick.profile`. We should be able to look at the profile choice and pick the right SequenceNextValUpdater implementation from that.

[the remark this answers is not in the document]
debasishg: @patriknw Now fixed. `schemaType` removed from `reference.conf`

debasishg:
Closing this PR after squashing all commits. This is going to be the feature branch for durable state implementation.

PR #550, opened 2021-06-30
Adds test for deleting events

nvollmar:
Adds test case for deleting events

Refs #551

### DIFF
--- a/core/src/test/scala/akka/persistence/jdbc/DeleteTest.scala
+++ b/core/src/test/scala/akka/persistence/jdbc/DeleteTest.scala
@@ -1,0 +1,124 @@
+package akka.persistence.jdbc
+
+import akka.actor.{ActorRef, ActorSystem, Props}
+import akka.persistence.{DeleteMessagesSuccess, PersistentActor}
+import akka.persistence.jdbc.query.scaladsl.JdbcReadJournal
+import akka.persistence.jdbc.testkit.javadsl.SchemaUtils
+import akka.persistence.query.{EventEnvelope, PersistenceQuery}
+import akka.testkit.{TestKit, TestProbe}
+import com.typesafe.config.ConfigFactory
+import org.scalatest.flatspec.AnyFlatSpecLike
+import org.scalatest.matchers.should.Matchers
+
+object DeleteTest {
+  val config: String =
+    """
+      |akka.actor.allow-java-serialization = on
+      |akka {
+      |  persistence {
+      |    journal {
+      |      plugin = "jdbc-journal"
+      |      // Enable the line below to automatically start the journal when the actorsystem is started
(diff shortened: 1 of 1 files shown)


core/src/test/scala/akka/persistence/jdbc/DeleteTest.scala
```diff
@@ -0,0 +1,124 @@
...
+    probe.send(persister, Persist("3"))
+    probe.expectMsg(AckPersist)
+
+    probe.send(persister, Delete(2))
+    probe.expectMsg(AckDelete(2)) // should delete
+
+    val eventProbe = TestProbe()
+    PersistenceQuery(system).readJournalFor[JdbcReadJournal]("jdbc-read-journal").eventsByPersistenceId("testId", 0, Long.MaxValue).runForeach(e => eventProbe.ref ! e)
+    val event = eventProbe.expectMsgType[EventEnvelope]
+    event.sequenceNr shouldBe 3 // succeeds with Cassandra or LevelDB, with H2 event seqNr 2 is returned
```
patriknw, line 122: Excellent finding. Can you add the fix to this PR?

nvollmar: The "fix" I found was just me poking around. 
I didn't go deep into the code of this project and honestly I have no idea if that is the correct way to fix it ;)

patriknw: I looked closer at this. There is already a filter in the `baseQuery`
https://github.com/akka/akka-persistence-jdbc/blob/068f0f0cc10db8fb38c4295bfa1397fd8e3ab87f/core/src/main/scala/akka/persistence/jdbc/query/dao/ReadJournalQueries.scala#L28-L29

Note the check of the config property for including deleted or not. By default it is using soft deletes, which was a surprise to me. I see that logicalDeletion has been deprecated but it is still true by default. Created issue https://github.com/akka/akka-persistence-jdbc/issues/557

Can you test again but set that config to false.

The purpose of keeping the last row is to know the latest sequenceNumber even when all events have been deleted. Sequence numbers should not be reused.

nvollmar: Thanks for the hint, I will retest it with that.

> The purpose of keeping the last row is to know the latest sequenceNumber even when all events have been deleted. Sequence numbers should not be reused.

The API should be consistent across different implementations (cassandra/jdbc). Since the Cassandra plugin allows deleting all events while maintaining the sequence number in a separate table, the JDBC plugin might follow the same pattern.

nvollmar: Right, that seems to get the correct behavior

PR #555, opened 2021-07-06
Initial implementation of new algo for re-query with delay in query

debasishg:
References https://github.com/akka/akka/issues/30277


core/src/main/scala/akka/persistence/jdbc/state/scaladsl/DurableStateSequenceActor.scala
```diff
@@ -83,104 +108,122 @@ class DurableStateSequenceActor[A](stateStore: JdbcDurableStateStore[A], config:
...
+      //    in order to find if it's still relevant
       // in this case we want to keep querying but not immediately
       scheduleQuery(queryDelay)
-      context.become(receive(nextMax, newMissingByCounter, (moduloCounter + 1) % maxTries))
+      context.become(receive(nextMax, missingOffsetCount + missingOffsetCurrentCount, (moduloCounter + 1) % maxTries))
+    }
+
+    // cache to be used in next iteration
+    if (!elements.isEmpty) {
+      elements.foreach(e => lastQueryResult += ((e.pid, e)))
```
patriknw, line 187: I think something is not optimal with updating this for each round, but I will have to play with some tests to see what happens.

patriknw (commented):
Looking good. A few cosmetic suggestions, and I will get back with more details when I have tried it with the test.

patriknw:
I tried this and it had a few deviations from my design idea so I started a new PR on top of this PR: https://github.com/akka/akka-persistence-jdbc/pull/556. Closing this because I think we should continue there.

PR #556, opened 2021-07-06
second iteration of algo for re-query with delay in query

patriknw:


core/src/main/scala/akka/persistence/jdbc/state/scaladsl/DurableStateSequenceActor.scala
```diff
@@ -167,20 +217,74 @@ class DurableStateSequenceActor[A](stateStore: JdbcDurableStateStore[A], config:
...
+        scheduleQuery(queryDelay)
+      }
+
+      // cache to be used in next iteration
+      if (elements.nonEmpty) {
+        elements.foreach(e => lastQueryResult += ((e.pid, e)))
+      }
+
+      context.become(receive(nextMax, newMissingByCounter, (moduloCounter + 1) % maxTries))
     } else {
```
patriknw, line 239: The first phase is using the same approach as before to find the offset gaps. Second phase if there are gaps is looking at the revisions...

core/src/test/scala/akka/persistence/jdbc/state/scaladsl/DurableStateSequenceActorTest.scala
```diff
@@ -58,83 +59,49 @@ abstract class DurableStateSequenceActorTest(config: Config, schemaType: SchemaT
...
 class MockDurableStateSequenceActorTest extends SharedActorSystemTestSpec {
-  def fetchMaxOrderingId(durableStateSequenceActor: ActorRef): Future[Long] = {
-    durableStateSequenceActor.ask(GetMaxOrderingId)(20.millis).mapTo[MaxOrderingId].map(_.maxOrdering)
+  def fetchMaxGlobalOffset(durableStateSequenceActor: ActorRef): Future[Long] = {
+    durableStateSequenceActor.ask(GetMaxGlobalOffset)(20.millis).mapTo[MaxGlobalOffset].map(_.maxOffset)
   }
 
-  it should "re-query with delay only when events are missing." in {
-    val batchSize = 100
+  it should "not delay several updates to known pid" in {
```
patriknw, line 66: difficult to see the diff here, but this test is new

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/DurableStateSequenceActor.scala
```diff
@@ -138,24 +188,24 @@ class DurableStateSequenceActor[A](stateStore: JdbcDurableStateStore[A], config:
...
-      elements.foldLeft[(OrderingId, OrderingId, MissingElements)](
-        (currentMaxOrdering, currentMaxOrdering, MissingElements.empty)) {
-        case ((currentMax, previousElement, missing), currentElement) =>
+      elements.foldLeft[(GlobalOffset, GlobalOffset, MissingElements)](
+        (currentMaxOffset, currentMaxOffset, MissingElements.empty)) {
+        case ((currentMax, previousOffset, missing), currentElement) =>
           // we must decide if we move the cursor forward
           val newMax =
-            if ((currentMax + 1).until(currentElement).forall(givenUp.contains)) {
+            if ((currentMax + 1).until(currentElement.offset).forall(givenUp.contains)) {
```
debasishg, line 196: @patriknw do we have to move the cursor one element at a time. In case of journal impl we used to store all the offsets in sequence - here we store the last offset only. Hence the strategy that I implemented was :

```
for all elements in the results
    if (total revision changes == total offset movement), then we have no gaps
```

This also takes into consideration the fact that we are using the READ COMMITTED isolation level and hence we cannot see upsert for revision `n+1` without seeing the upsert for revision `n`

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/DurableStateSequenceActor.scala
debasishg: @patriknw do we need the first phase ? Will not the revisions give us the full information ?

debasishg: @patriknw The gaps that you will find in the first phase will be found from the revisions + offset information as well (second phase). The difference is now, in the second phase, you have the math to verify that those are not real gaps.

patriknw: I think the problem is for following rounds.

Also, we should cover the scenarios of gaps because of sequence visibility and transaction rollback in the same way as before, i.e. query again 10 times.

Anyway, this is difficult enough that it must be driven by tests. First step would be to add back some of the original tests. I can continue with that now.

debasishg: > gaps because of sequence visibility and transaction rollback in the same way as before, i.e. query again 10 times.

These are real gaps and would be visible in Phase 2 as well because the revision math will not align. And then re-query will follow.

> Anyway, this is difficult enough that it must be driven by tests.

+1

core/src/main/scala/akka/persistence/jdbc/state/scaladsl/DurableStateSequenceActor.scala
```diff
@@ -167,20 +217,78 @@ class DurableStateSequenceActor[A](stateStore: JdbcDurableStateStore[A], config:
...
+        self ! QueryState
+      } else {
+        // keep querying but not immediately
+        scheduleQuery(queryDelay)
+      }
+
+      // cache to be used in next iteration
+      if (elements.nonEmpty) {
+        elements.foreach(e => revisionCache += ((e.pid, e)))
+      }
```
debasishg, line 237: @patriknw we don't need this - right ? We are calling the function to populate the cache in the next line ..

patriknw: thanks, it was an accidental leftover

patriknw:
@debasishg This is now in a good shape and ready for final review, if you can take a look.

Let me squash my commits before we merge to the feature branch.

debasishg (approved):
LGTM

patriknw:
@debasishg Tests are flaky when running in CI. We have very low timeouts in some places, like 150 to 200 millis. That's typically not working well in CI environments. It's be more like seconds. In Akka we try to not have test timeouts shorter than 3 seconds.

I'm merging this anyway because I don't have the time to investigate now. We can follow up on that separately.

Issue #557, opened 2021-07-08
Change the default for akka-persistence-jdbc.logicalDeletion.enable


patriknw:
Soft (logical) deletes is still the default. I can imagine that is surprising for many. Logical deletes has been deprecated and there is a warning logged.

ignasi35:
I haven't jumped into this codebase in a while and realized that `logicalDelete` is only used in the (now legacy) `ByteArrayJournalDao`. The `ByteArrayJournalDao` class is not deprecated but was moved to the `akka.persistence.jdbc.journal.dao.legacy` package prior to version 5.0.0. See https://github.com/akka/akka-persistence-jdbc/pull/467

If I am reading the code and history correctly, in v5.0.0 the new `DefaultJournalDao` was introduced and that new DAO doesn't support logicalDelete.

Then, if introducing a change like switching the default value of `akka-persistence-jdbc.logicalDeletion.enable` from `false` to `true` already requires a MAJOR version bump, I wonder if we could, instead, drop the `ByteArrayJournalDao` and require users to migrate their data and use `DefaultJournalDao`.

octonato:
> require users to migrate their data

That's the problem. We don't have a migration tool yet. That's why we're keep the legacy DAO for now. 

The is a migration tool in the making (#501). I was even planning to pick it this week, but didn't find the time.

ignasi35:
Sure, what I'm saying is that in a (hypothetical) version v6.0.0 of this plugin we would have the migration tool (from the legacy `ByteArrayJournalDao` to the `DefaultJournalDao`). Then, changing the default value of `logicalDeletion.enable` is no longer valuable. Users gain a lot more from migrating than from adopting the new default value of `logicalDeletion.enable` and staying on `ByteArrayJournalDao`.
I would just deprecate `ByteArrayJournalDao` in `v5.0.x` already or add a log trace linking to an issue explaining the plans to remove it when the migration tool is completed.

IMHO; any improvement on `ByteArrayJournalDao` is little to no value.

PR #569, opened 2021-07-23
Removes logical deletion

nvollmar:
References #557


ignasi35:
Thanks @nvollmar for puting this in motion.

This is a rather important behavioral change so, IMHO it should be prominently explained on the docs and release notes. Maybe even justifies the bump of the MAJOR version of the plugin (as a user I'd be very pissed if a PATCH release included such a change sneakily).

Would you be able to review the docs for places where delete is mentioned and amend them? Thanks!

ignasi35:
In general, SQL queries must still filter by `logicalDeletion.enable = false` but maybe when the flag to hard-delete is enabled we can remove the filter by `journal.deleted=false` criteria from the SLQ statement. Not on this PR though, that could be a future improvement.

ignasi35:
Do we already have documentation on the safest procedure to move from `logicalDeletion.enable = false` to `logicalDeletion.enable = true` (and back in case of a rollback)? Maybe for a separate PR too.

nvollmar:
I agree that this would be quite a significant change that should be mentioned prominently in the release notes. As mentioned in Issue #557, this feature is deprecated and warned. If a major release is warranted, then the feature might also be removed at the same time to clean everything up.

I didn't find any relevant mentioning of deletion in the docs.

nvollmar:
There is no mentioning of logical deletion anywhere in the `docs` folder of the project

ignasi35:
Thanks for checking.

Maybe, since there's already a deprecation warning and changing the default value of the setting already requires a MAJOR version bump, maybe we could just drop the setting and the feature altogether so users migrating to that major version will have to adopt the new behavior.

nvollmar:
I've added the removal of this feature to this PR

ignasi35 (commented):
Nice! thanks @nvollmar

octonato:
@nvollmar, we will need MiMa exclusions.

nvollmar:
@octonato Added mima excludes

patriknw (approved):
LGTM

patriknw:
@nvollmar sorry for the delay here. Would you be able to fix the conflicts? I think this is ready to be merged after that.

patriknw:
@nvollmar `OracleScalaEventAdapterTest` is failing here. Do you think it can be related to the changes in this PR? I have re-run it at least twice.

Also ran it successfully on master in https://github.com/akka/akka-persistence-jdbc/pull/672

patriknw:
Note that the failing job is running with some custom parameters:
```
sbt "++2.13.4 It/testOnly akka.persistence.jdbc.integration.Oracle*"   -Djdbc-journal.dao=akka.persistence.jdbc.journal.dao.legacy.ByteArrayJournalDao -Djdbc-snapshot-store.dao=akka.persistence.jdbc.snapshot.dao.legacy.ByteArraySnapshotDao -Djdbc-read-journal.dao=akka.persistence.jdbc.query.dao.legacy.ByteArrayReadJournalDao
```

nvollmar:
I'll have to backtrack the changes and see what might cause this

patriknw:
I'm quite convinced that the Oracle case in akka.persistence.jdbc.query.dao.legacy.ByteArrayReadJournalDao was broken already before this change when using logicalDelete=false
That shows up now when the default has been changed.
I verified that in https://github.com/akka/akka-persistence-jdbc/pull/672

I will merge this and then mark the failing tests as pending in a separate PR. I don't think we care that much about the legacy.ByteArrayReadJournalDao

PR #571, opened 2021-07-28
Consistent syntax postgres-create-schema.sql

ignasi35:

PR #572, opened 2021-07-29
Initial implementation of jdbc based DurableStateStore and query

ignasi35:
References #571


ignasi35:
Thi sis not working out the way I anticipated. The `state` table name is hardcoded in several places even though it should be read from `reference.conf`.

Another problem making this refactor bigger than I anticipated is that we're using `sqlu` (slick query string interpolation) which doesn't like dynamic table and column names. When I rewrite the SQLu interpolation to use dynamic table and column names:

```
    sqlu"""INSERT INTO ${durableStateTableCfg.tableName}
            (
              ${durableStateTableCfg.columnNames.persistenceId},
              ${durableStateTableCfg.columnNames.globalOffset},
              ${durableStateTableCfg.columnNames.revision},
              ${durableStateTableCfg.columnNames.statePayload},
              ${durableStateTableCfg.columnNames.stateSerId},
              ${durableStateTableCfg.columnNames.stateSerManifest},
              ${durableStateTableCfg.columnNames.tag},
              ${durableStateTableCfg.columnNames.stateTimestamp}
            )
            VALUES
            (
              ${row.persistenceId},
              #${seqNextValue},
              ${row.revision},
              ${row.statePayload},
              ${row.stateSerId},
              ${row.stateSerManifest},
              ${row.tag},
              #${System.currentTimeMillis()}
            )
      """
```

Slick considered them to be arguments on a prepared statement and produced the following SQL statement: 

```

[info] INSERT INTO ?
[info]             (
[info]               ?,
[info]               ?,
[info]               ?,
[info]               ?,
[info]               ?,
[info]               ?,
[info]               ?,
[info]               ?
[info]             )
[info]             VALUES
[info]             (
[info]               ?,
[info]               NEXT VALUE FOR "PUBLIC"."SYSTEM_SEQUENCE_F55EBA0E_9EDF_41B5_BB28_75006EFBEAFB",
[info]               ?,
[info]               ?,
[info]               ?,
[info]               ?,
[info]               ?,
[info]               1627547135755
[info]             )
```

I need to find a way for this to work or we'll have to migrate to using Slick API (as we do in the rest of the plugin).

ignasi35:
For the time being, I'll _just_ hardcode `durable_state` everywhere and open an issue to migrate the code in a separate PR.

debasishg:
@ignasi35 regarding using slick API for the insert, I could not make it work because of the sequence thing

debasishg:
@ignasi35 do we need to change the name of the table right now ? We can do it in a future PR - right ?

debasishg (approved):
Changing the table name LGTM

ignasi35:
> @ignasi35 regarding using slick API for the insert, I could not make it work because of the sequence thing

I wonder how that works on the `event_journal` table (if at all 😒). Let's discuss in #573 

BTW: I'm 100% in favor of getting something working first and configurable later. That's why I prefer expediting this PR and working on #573 later.

Issue #573, opened 2021-07-29
Make durable-state SQL statements configurable


ignasi35:
See https://github.com/akka/akka-persistence-jdbc/pull/572#issuecomment-888919506

There's a few usages of hardcoded SQL that should, instead, read the table and column names from `reference.conf`. See for example the following INSERT statement:

```scala
  private[jdbc] def insertDbWithDurableState(row: DurableStateTables.DurableStateRow, seqNextValue: String) = {
    // TODO: read the table name and column names from durableStateTableCfg
    sqlu"""INSERT INTO durable_state
            (
              persistence_id,
              global_offset,
              revision,
              state_payload,
              state_serial_id,
              state_serial_manifest,
              tag,
```

Instead, it should read value from `reference.conf`:

```
  tables {
    durable_state {
      ## The table and column names are not always read and used in SQL statements. If you change
      ## these values you may need to edit some source code
      ## https://github.com/akka/akka-persistence-jdbc/issues/573
      tableName = "durable_state"
      schemaName = ""
      columnNames {
        globalOffset = "global_offset"
        persistenceId = "persistence_id"
        revision = "revision"
        statePayload = "state_payload"
        stateSerId = "state_serial_id"
        stateSerManifest = "state_serial_manifest"
        tag = "tag"
        stateTimestamp = "state_timestamp"
      }
    }
  }
```

ignasi35:
done in #575

PR #575, opened 2021-07-29
Durable state DDL is all configurable

ignasi35:
References #573 

Thanks @debasishg for pointing out the correct syntax (`#$`). TIL

PS: I've tested locally both PG and H2.


debasishg (approved):
LGTM

Issue #540, opened 2021-05-24
Failed: H2ScalaEventsByTagTest


johanandren:
https://github.com/akka/akka-persistence-jdbc/runs/2656339029?check_suite_focus=true

```
[info] - should find all events by tag even when lots of events are persisted concurrently *** FAILED *** (1 minute)
662
[info]   java.lang.AssertionError: Expected OnNext(_), yet no element signaled during 58297543317 nanoseconds
663
[info]   at akka.stream.testkit.TestSubscriber$ManualProbe.expectNext(StreamTestKit.scala:368)
664
[info]   at akka.stream.testkit.TestSubscriber$ManualProbe.expectNext(StreamTestKit.scala:359)
665
[info]   at akka.persistence.jdbc.query.EventsByTagTest.$anonfun$new$35(EventsByTagTest.scala:214)
666
[info]   at akka.persistence.jdbc.query.EventsByTagTest.$anonfun$new$35$adapted(EventsByTagTest.scala:213)
667
[info]   at scala.collection.TraversableOnce$folder$1.apply(TraversableOnce.scala:196)
668
[info]   at scala.collection.TraversableOnce$folder$1.apply(TraversableOnce.scala:194)
669
[info]   at scala.collection.immutable.Range.foreach(Range.scala:158)
670
[info]   at scala.collection.TraversableOnce.foldLeft(TraversableOnce.scala:199)
671
[info]   at scala.collection.TraversableOnce.foldLeft$(TraversableOnce.scala:192)
672
[info]   at scala.collection.AbstractTraversable.foldLeft(Traversable.scala:108)
673
[info]   at akka.persistence.jdbc.query.EventsByTagTest.$anonfun$new$34(EventsByTagTest.scala:213)
674
[info]   at akka.persistence.jdbc.query.EventsByTagTest.$anonfun$new$34$adapted(EventsByTagTest.scala:211)
```

johanandren:
https://github.com/akka/akka-persistence-jdbc/runs/3410478161?check_suite_focus=true#step:6:783

johanandren:
https://github.com/akka/akka-persistence-jdbc/pull/600/checks?check_run_id=3645817404

johanandren:
https://github.com/akka/akka-persistence-jdbc/pull/600/checks?check_run_id=3645817436

johanandren:
https://github.com/akka/akka-persistence-jdbc/pull/595/checks?check_run_id=3645812333

[...]

PR #601, opened 2021-09-27
fix: fail fast in query tests if actors fail to stop 

johanandren:
References #540

I have a suspicion about the H2ScalaEventsByTagTest failures being
about journal init, but not sure. This will help pinpointing that.


core/src/test/scala/akka/persistence/jdbc/query/EventsByPersistenceIdTest.scala
```diff
@@ -171,8 +171,8 @@ abstract class EventsByPersistenceIdTest(config: String) extends QueryTestSpec(c
 
   it should "deliver EventEnvelopes non-zero timestamps" in withActorSystem { implicit system =>
     val journalOps = new ScalaJdbcReadJournalOperations(system)
+    val testStartTime = System.currentTimeMillis()
     withTestActors(replyToMessages = true) { (actor1, actor2, actor3) =>
-      val testStartTime = System.currentTimeMillis()
```
johanandren, line 175: This was a race before this fix as well, but the forced ping-pong with each actor makes it fail consistently each time now since actors are _always_ started when we get here now.

patriknw (approved):
LGTM

PR #603, opened 2021-10-11
Migration tool

octonato:
This is a feature branch with initial work done by @Tochemey in #501. 

I will keep it as draft as until branch `migration-tool` is ready to be merged into `master`.


octonato:
@AlixBa, when we deserialize `PersistentRepr`, the payload gets automatically deserialized. That's how `akka.persistence.serialization.MessageSerializer` does it and we are just re-using this serializer. 
But I agree that strictly speaking we don't need to deserialize the payload. 

On the other hand, it's a good sanity check as it reveals if the journal is readable or not. 

I'm not sure if the best solution is to make the migration oblivious about the payload. It's good to just rely on the existing serializer so at least we are sure that we are reading the data in the same way it is expected to be read in a production environment. 

Eventually, we could provide two migration method. One using the MessageSerializer by default and one that the user could pass its own flow. Then people could eventually drop events, transform, etc.

patriknw (approved):
LGTM (reviewed briefly)
Thanks to the contributors!

Issue #650, opened 2022-03-25
SQL Server deadlocks during recovery/restart


JustinPihony:
### Versions used 

Akka version: 2.6.13
Akka Persistence JDBC: 5.0.4
mssql-jdbc: 9.5.0.jre8-preview

Remember Entities is also being used

### Expected Behavior

The application should recover and restart any entities without exceptions

### Actual Behavior

Upon restarting (normally or through recovery) and an old entity is restarted then a deadlock is encountered.

### Relevant logs

RecoveryFailed(com.microsoft.sqlserver.jdbc.SQLServerException: Transaction (Process ID 538) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction.)

[ERROR] [...Entity] [] [...actor.default-dispatcher-20] - Supervisor RestartSupervisor saw failure: Exception during recovery. Last known sequence number [0]. PersistenceId [...], due to: Exception during recovery. Last known sequence number [0]. PersistenceId [...], due to: Transaction (Process ID 209) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction.
akka.persistence.typed.internal.JournalFailureException: Exception during recovery. Last known sequence number [0]. PersistenceId [...], due to: Exception during recovery. Last known sequence number [0]. PersistenceId [...], due to: Transaction (Process ID 209) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction.

An example deadlocked query set:

`(@P0 nvarchar(4000))select max(x2.x3) from (select top (1) "persistence_id" as x4, "sequence_number" as x3 from "event_journal" where "persistence_id" = @P0 order by "sequence_number" desc) x2`

AND

`@P0 bit,@P1 nvarchar(4000),@P2 bigint,@P3 nvarchar(4000),@P4 bigint,@P5 nvarchar(4000),@P6 varbinary(max),@P7 int,@P8 nvarchar(4000),@P9 varbinary(max),@P10 int,@P11 nvarchar(4000))insert into "event_journal" ("deleted","persistence_id","sequence_number","writer","write_timestamp","adapter_manifest","event_payload","event_ser_id","event_ser_manifest","meta_payload","meta_ser_id","meta_ser_manifest")  values (@P0,@P1,@P2,@P3,@P4,@P5,@P6,@P7,@P8,@P9,@P10,@P11) select SCOPE_IDENTITY() AS GENERATED_KEYS`

[...]

PR #652, opened 2022-03-28
migrate SQL Server Schemas to NVARCHAR

leviramsey:
References #650

### DIFF
--- a/core/src/main/resources/schema/sqlserver/sqlserver-create-schema-varchar.sql
+++ b/core/src/main/resources/schema/sqlserver/sqlserver-create-schema-varchar.sql
@@ -1,5 +1,5 @@
 /*
-Akka Persistence JDBC versions from 5.0.0 through 5.0.4 used this schema.  The only difference from the
+Akka Persistence JDBC versions from 5.0.0 through 5.1.0 used this schema.  The only difference from the
 post-5.0.4 schema is the use of VARCHAR instead of NVARCHAR for string fields.  It is strongly
 recommended that new uses of Akka Persistence JDBC 5.0.0 and later use the NVARCHAR schema.  This schema is
 still usable with post-5.0.4 versions of Akka Persistence JDBC, though will not support Unicode persistence IDs,


leviramsey:
Re: docs, I'm not sure where a "if using a journal/snapshot store created with Akka Persistence versions 5.0.4 and earlier, it is advised to disable the JDBC client's sending strings as unicode" note should go.

leviramsey:
Thoughts on restructuring the docs along the lines of the Cassandra plugin with notes for the different DBs?

octonato:
About the docs, I think we should add it to the migration guide. 

Now it's a good moment to do it. We have another [PR](https://github.com/akka/akka-persistence-jdbc/pull/648), less critical, that will also require a migration guide. 

And we didn't documented the migration tool that was added in #603. 

Maybe we should go for 5.1.0 because of the DB changes.

leviramsey:
We should document a recommendation if using the old (`VARCHAR`-based) SQL Server schemas to use the `;sendStringParametersAsUnicode=false` connection parameter.

leviramsey:
Actually I guess we can support offline migration of data from an old journal/snapshot-store to a new journal/snapshot-store, but I think it would be generally preferable from the user perspective to use the old schema and configure JDBC to work properly with the old schema.

That said, the migration tool would support the "turns out, we need unicode persistence IDs or tags" use case.

patriknw:
@leviramsey would it be possible to push this forward?

leviramsey:
I guess the question is if we want to implement a migration tool (which honestly is not something I'm confident on being able to implement any time soon).

As it is, this should just be mergeable (it requires no code changes, AFAIK), perhaps with docs around `;sendStringParametersAsUnicode=false` for use with previous versions of the schema.

core/src/main/resources/schema/sqlserver/sqlserver-create-schema-varchar.sql
```diff
@@ -0,0 +1,53 @@
+/*
+Akka Persistence JDBC versions from 5.0.0 through 5.0.4 used this schema.  The only difference from the
```
patriknw, line 2: also 5.1.0

patriknw (approved):
Looking good. We don't need to provide a migration tool. DBAs (or any developer with some google skills) can do that with SQL.

Only a small addition of the version numbers.

johanandren (approved):

Issue #639, opened 2022-02-14
SchemaUtils.createIfNotExists(.) doesn't work with H2 Version 2.1.210


alphaville76:

octonato:
That's correct. You should indeed use `SchemaUtils.applyScript()`. 

Beware that the plugin is not tested with H2 Version 2, so I hope you won't run into any other issues.

alphaville76:
Thanks for the reply, but you should update or drop the H2 v.1 support because of a Log4Shell-Like Vulnerability in the Version 1: https://www.securityweek.com/log4shell-vulnerability-found-popular-h2-database

See also:
[CVE-2022-23221](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-23221)
[CVE-2021-42392](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-42392)
[CVE-2021-23463](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-23463)

I let you know, if I run into other issues.

alphaville76:
For the moment, I'm using this script (very similar to the versioned one):

CREATE TABLE IF NOT EXISTS "event_journal"
(
    "ordering"           BIGINT                NOT NULL AUTO_INCREMENT,
    "deleted"            BOOLEAN DEFAULT false NOT NULL,
    "persistence_id"     VARCHAR(255)          NOT NULL,
    "sequence_number"    BIGINT                NOT NULL,
    "writer"             VARCHAR               NOT NULL,
    "write_timestamp"    BIGINT                NOT NULL,
    "adapter_manifest"   VARCHAR               NOT NULL,
    "event_payload"      BLOB                  NOT NULL,
    "event_ser_id"       INTEGER               NOT NULL,
    "event_ser_manifest" VARCHAR               NOT NULL,
    "meta_payload"       BLOB,
    "meta_ser_id"        INTEGER,
    "meta_ser_manifest"  VARCHAR,
    PRIMARY KEY ("persistence_id", "sequence_number")
);

CREATE UNIQUE INDEX IF NOT EXISTS "event_journal_ordering_idx" on "event_journal" ("ordering");

CREATE TABLE IF NOT EXISTS "event_tag"
(
    "event_id" BIGINT  NOT NULL,
    "tag"      VARCHAR NOT NULL,
    PRIMARY KEY ("event_id", "tag")
);

CREATE TABLE IF NOT EXISTS "snapshot"
(
[...]

PR #656, opened 2022-06-08
Add migration notes for H2 upgrade

nvollmar:
References #648 #639

Trying to get things rolling as we (and probably others) are waiting on this.

### DIFF
--- a/docs/src/main/paradox/migration.md
+++ b/docs/src/main/paradox/migration.md
@@ -1,5 +1,13 @@
 # Migration
 
+## Migrating to version 6.0.0
+
+**Release `6.0.0` updates H2 to version 2.1.212 which is not compatible to the previous 1.4.200***
+
+H2 has undergone considerable changes that broke backwards compatibility to make H2 SQL Standard compliant.
+For migration please refer to the H2 [migration guide](http://www.h2database.com/html/migration-to-v2.html)
+
+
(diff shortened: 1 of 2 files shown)


octonato (commented):
Looking good. 

Could you also add a mention to the documentation that we don't recommend H2 to be used as a production database?

octonato:
@nvollmar, we can cut RC if you are waiting for this release. Let me know if needed.

nvollmar:
It is not a big deal, but it causes warning on security checks of dependencies. Just wanted to keep things rolling to get that fixed eventually :)

README.md
```diff
@@ -7,6 +7,8 @@
 akka-persistence-jdbc writes journal and snapshot entries to a configured JDBC store. It implements the full akka-persistence-query API and is therefore very useful for implementing DDD-style 
 application models using Akka for creating reactive applications.
 
+Please note that the H2 database is not recommended to be used as a production database, and support for H2 is primarily for testing purposes.
+
```
octonato, line 11: Could we have that same sentence in that page? 
https://doc.akka.io/docs/akka-persistence-jdbc/current/configuration.html 

We can use a callout just after the list of supported databases: 
https://developer.lightbend.com/docs/paradox/current/directives/callouts.html

octonato (approved):
Thanks @nvollmar.

PR #681, opened 2022-09-16
Limit events by tag query ordering sizes

johanandren:
This makes it possible to limit the chunks of the journal queried for events by tag to work around bad performance in some dbs when starting a new query/projection from 0 after a lot of tagged events was already persisted. By default the limiting is disabled, will need app specific tuning when enabled.

### DIFF
--- a/core/src/main/scala/akka/persistence/jdbc/config/AkkaPersistenceConfig.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/config/AkkaPersistenceConfig.scala
@@ -188,6 +188,7 @@ class ReadJournalConfig(config: Config) {
   val refreshInterval: FiniteDuration = config.asFiniteDuration("refresh-interval")
   val maxBufferSize: Int = config.getInt("max-buffer-size")
   val eventsByTagBufferSizesPerQuery: Long = config.getLong("events-by-tag-buffer-sizes-per-query")
+  require(eventsByTagBufferSizesPerQuery >= 0, "events-by-tag-buffer-sizes-per-query must not be negative")
   val addShutdownHook: Boolean = config.getBoolean("add-shutdown-hook")
 
   override def toString: String =


lightbend-cla-validator:
Hi @JustinPihony,

We see that you have signed the Lightbend Contributors License Agreement before, however, the CLA has changed since you last signed it.
Please review the new CLA and sign it before we proceed with reviewing this pull request:

johanandren:
Tested with a `events-by-tag-buffer-sizes-per-query = 10`, all tests green. I'll revert it to default as disabled now.

core/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -228,15 +228,29 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
     import FlowControl._
     implicit val askTimeout: Timeout = Timeout(readJournalConfig.journalSequenceRetrievalConfiguration.askTimeout)
     val batchSize = readJournalConfig.maxBufferSize
+    val maxOrderingRange = readJournalConfig.eventsByTagBufferSizesPerQuery match {
+      case 0 => None
+      case x => Some(x * batchSize)
+    }
+
+    def getLoopMaxOrderingId(offset: Long, latestOrdering: MaxOrderingId): Future[MaxOrderingId] =
```
patriknw, line 236: why is this a Future if it's always `Future.successful`?

core/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -228,15 +228,29 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
...
+      })
 
     Source
       .unfoldAsync[(Long, FlowControl), Seq[EventEnvelope]]((offset, Continue)) { case (from, control) =>
         def retrieveNextBatch() = {
           for {
             queryUntil <- journalSequenceActor.ask(GetMaxOrderingId).mapTo[MaxOrderingId]
-            xs <- currentJournalEventsByTag(tag, from, batchSize, queryUntil).runWith(Sink.seq)
+            loopMaxOrderingId <- getLoopMaxOrderingId(from, queryUntil)
+            xs <- currentJournalEventsByTag(tag, from, batchSize, loopMaxOrderingId).runWith(Sink.seq)
```
johanandren, line 251: Each fetch is still limited by batch size, but the far end of the query to consider is capped by `from + maxOrderingRange` elements, for unevenly tagged event streams that could lead to many empty results, so user will have to benchmark and tune in their specific journal.

core/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -228,15 +228,29 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
...
           } yield {
-            val hasMoreEvents = xs.size == batchSize
+            val hasMoreEvents = (xs.size == batchSize) || (loopMaxOrderingId.maxOrdering < queryUntil.maxOrdering)
```
patriknw, line 253: doesn't make it easier to understand the end conditions here, but fine I guess

patriknw (approved):
LGTM

core/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
johanandren: Good point, it's used in a for-comprehension on future but that shouldn't be needed. I'll fix.

core/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
johanandren: I'll add a comment, that can help a little at least.

JustinPihony (approved):
LGTM

PR #687, opened 2022-09-22
Don't publish release twice

patriknw:
* both were triggered by tag
* snapshots are anyway published for master on each commit


johanandren:
Some docs referring to the snapshot branch in RELEASING or if it was README as well.

johanandren (approved):

PR #688, opened 2022-09-23
fix: The new events-by-tag-buffer-sizes-per-query could miss events

johanandren:
for infrequent tags because of an optimization that only works for full journal querying

Follow up to #681


sebastian-alfers:
Thanks @johanandren for kicking this off. The change you initially proposed (`xs.map(_.offset.value).max`) did cause a `java.lang.UnsupportedOperationException: empty.max` as at some point `xs` was empty.

However, I think the bug was in the wrong usage of `nextStartingOffset` for the "next" batch not considering the config in case it exists (recently introduced by `loopMaxOrderingId`).

JustinPihony (approved):

johanandren:
Indeed. LGTM (can't mark with approve since it was my initial PR)

johanandren:
Squashed and attributed to Sebastian.

sebastian-alfers (approved):

PR #719, opened 2023-02-12
Support schema in upsert for durable store

cwholmes:
The durable state store has configuration for a schema of the durable_state table. This configuration seems to be supported for the get and delete object methods, but not for the upsert method. This PR adds support for that configuration when upserting.

This is an addition to changes in #575

### DIFF
--- a/core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
@@ -47,7 +47,7 @@ import akka.persistence.jdbc.config.DurableStateTableConfiguration
     durableStateTable.filter(_.persistenceId === persistenceId)
 
   private[jdbc] def insertDbWithDurableState(row: DurableStateTables.DurableStateRow, seqNextValue: String) = {
-    sqlu"""INSERT INTO #${durableStateTableCfg.tableName}
+    sqlu"""INSERT INTO #${durableStateTableCfg.schemaName}.#${durableStateTableCfg.tableName}
             (
              #${durableStateTableCfg.columnNames.persistenceId},
              #${durableStateTableCfg.columnNames.globalOffset},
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
Hi @cwholmes,

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
```diff
@@ -47,43 +47,36 @@ import akka.persistence.jdbc.config.DurableStateTableConfiguration
...
+        (
+          seqNextValue,
+          row.revision,
+          row.statePayload,
+          row.stateSerId,
+          row.stateSerManifest,
+          row.tag,
+          System.currentTimeMillis()
+        )
+      )
```
patriknw, line 79: I'm quite sure there was a reason why we used the string based statements.

I think the schema is optional so you can probably construct it with something like 
```
schema.map(_ + ".").getOrElse("") + table
```

cwholmes: I know that the current implementation won't work as the `seqNextVal` is a string, but the global offset is a long (which would need to be addressed). I will try to revert to the string statements, but would be interested as to the the reasoning for these statements since the resulting sql should be much the same.

The issue using `.map.getOrElse` is the interpolation and requirement of `#` in some cases that wouldn't allow a blank schema, but I will give it a try.

cwholmes: I have restored the string sql statements and updated it such that it should maintain the `schema.table` literal created from the fold statement on the optional schema.

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
```diff
@@ -47,7 +47,9 @@ import akka.persistence.jdbc.config.DurableStateTableConfiguration
     durableStateTable.filter(_.persistenceId === persistenceId)
 
   private[jdbc] def insertDbWithDurableState(row: DurableStateTables.DurableStateRow, seqNextValue: String) = {
-    sqlu"""INSERT INTO #${durableStateTableCfg.tableName}
+    val tableAndSchema = durableStateTableCfg.schemaName.fold(durableStateTableCfg.tableName)(schema =>
```
patriknw, line 50: can this be moved to an instance `val` of the class to avoid this for each call?

patriknw (commented):
looking good

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
cwholmes: Moved to an instance val.

johanandren (approved):
LGTM, thanks!

patriknw (approved):
LGTM, thanks

PR #727, opened 2023-02-21
build: Fix release tag pattern

johanandren:


.github/workflows/release.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches:
       - main
       - master
-    tags: ["v.*"]
+    tags: ["v*"]
```
patriknw, line 8: in some other places we even have `tags: [ "*" ]`

ennru: Tags might be useful for other things than releases.

ennru (approved):
LGTM.

Issue #710, opened 2022-11-22
batch insert event and event_tag performance.


Roiocam:
Recently I am doing load testing on my application. it does not really have a good performance, and even can not reach the bottleneck of the database.

I have some questions on the insert model:

on `akka.persistence.jdbc.journal.dao.DefaultJournalDao`, it uses multi insert action `++=`

https://github.com/akka/akka-persistence-jdbc/blob/bba266cc0b0d398e0b16fa0d1347311f51116b48/core/src/main/scala/akka/persistence/jdbc/journal/dao/DefaultJournalDao.scala#L45

And then I took some time to find the source that executes SQL. The multi-insert action implement is `MultiInsertAction` on [`slick.jdbc.JdbcActionComponent`](
https://github.com/slick/slick/blob/4477219e4090620d1c6dcb26bf005b1e2158ac26/slick/src/main/scala/slick/jdbc/JdbcActionComponent.scala#L525
)

And it basically just uses `java.sql.Statement#executeBatch`. as far as I know it wouldn't optimize too much on the multi-insert on this method. (Most of the JDBC drivers would not). 

They just for-each loop every param and fill them to statement and then for-each loop executes it, it will take huge network IO (the bottleneck).

MySQL document points out how to optimize this.  https://dev.mysql.com/doc/refman/8.0/en/insert-optimization.html

General, most database support insert values list, it is just one network i/o.

I believe it is a performance issue, and break down the server when the network was degraded. (Expressed as a lot of EventSourcedBehavior complain circuit breakers fail fast).

In the end, In the process of writing down this issue (collecting information), I found the same issue already exists, but on scala slick.

https://github.com/slick/slick/pull/2398 and https://github.com/slick/slick/issues/1272

Roiocam:
i think this was related on #592

Roiocam:
> @Roiocam, thanks for reporting this.
> 
> I'm curious to see the results of the same load test after adding the PK/FK as described in #592.

yep, i'm working on it.

I fork this repository and change the source code base on #592, packaged it and uploaded it to our Maven repository (I am not familiar with scala, sbt publish blocked me most of the time, and finally solved it through maven deploy-file)

I am using same load test for this, but has more case:

- **CASE 1**
	- `hardware:` 4c8g (2 of node with profiler)
	- `persit actor num:` 1000
	- `throughput:` 750
-  **CASE 2**
	- `hardware:` 4c8g (2 of node with profiler)
[...]

PR #731, opened 2023-04-13
Replace event_tag FK to get rid of insert and return #710

Roiocam:
References #710


patriknw (commented):
looking good, but a compatibility approach is needed

core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
```diff
@@ -20,22 +20,21 @@ class JournalQueries(
...
 
-  def writeJournalRows(xs: Seq[(JournalAkkaSerializationRow, Set[String])])(implicit ec: ExecutionContext) = {
+  def writeJournalRows(xs: Seq[(JournalAkkaSerializationRow, Set[String])])(implicit ec: ExecutionContext): DBIOAction[Any, NoStream, Effect.Write] = {
     val sorted = xs.sortBy((event => event._1.sequenceNumber))
+    // not matter what, always insert event.
+    val events = sorted.map(_._1)
     if (sorted.exists(_._2.nonEmpty)) {
       // only if there are any tags
-      val (events, tags) = sorted.unzip
+      val tagsRows = sorted.flatMap(e => e._2.map(v => TagRow(e._1.persistenceId, e._1.sequenceNumber, v)))
```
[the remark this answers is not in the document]
Roiocam: Thanks

core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
```diff
@@ -20,22 +20,21 @@ class JournalQueries(
...
       for {
-        ids <- insertAndReturn ++= events
-        tagInserts = ids.zip(tags).flatMap { case (id, tags) => tags.map(tag => TagRow(id, tag)) }
-        _ <- TagTableC ++= tagInserts
+        _ <- JournalTableC ++= events
+        _ <- TagTableC ++= tagsRows
```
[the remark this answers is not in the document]
Roiocam: It looks like the configuration requires two flags.

And I am unable to locate the unit test for JournalQueries. Perhaps it is hidden within the integration test or there is some other detail that I may have missed.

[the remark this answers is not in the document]
Roiocam: I have made some updates to this. I have used two configuration property to achieve 'Writing both old and new' and `reading and switching between the new and old tables`,  what do you think about those change? @patriknw 

And i think the following steps could achieve the rolling update goal:

1. Use `writeLegacyTag=true` and `readLegacyTag=true`  to write and read using the old method, and write extra tags into the new table.
2. Once the projection consumer reaches the head of the new table, switch to the new reading method by setting `readLegacyTag=false`, but continue writing using the old method (Do we need to write data using both the old and new methods at this step?).
3. Finally, when the old table is no longer needed, use `writeLegacyTag=false` to complete the migration."


**write**

```scala
  private def writeEventsAndTags(sorted: Seq[(JournalAkkaSerializationRow, Set[String])])(
      implicit ec: ExecutionContext): DBIOAction[Any, NoStream, Effect.Write] = {
    val (events, tags) = sorted.unzip
    // for new table write
    val tagRows = sorted.flatMap { case (eventRow, tags) =>
      tags.map(t => TagRow(eventRow.persistenceId, eventRow.sequenceNumber, t))
    }
    // using old way write both table.
    if (tagTableCfg.writeLegacyTag) {
      for {
        ids <- insertAndReturn ++= events
        tagInserts = ids.zip(tags).flatMap { case (id, tags) => tags.map(tag => LegacyTagRow(id, tag)) }
        _ <- LegacyTagTableC ++= tagInserts
        _ <- TagTableC ++= tagRows
      } yield ()
    } else {
      // optimization using batch insert
      for {
        _ <- JournalTableC ++= events
        _ <- TagTableC ++= tagRows
      } yield ()
    }
  }
```

**read**

```scala
    // using old way read old table
    if (tagTableCfg.readLegacyTag) {
      baseTableQuery().join(LegacyTagTable).on(_.ordering === _.eventId)
    } else {
      // switch to new way
      baseTableQuery()
        .join(TagTable)
        .on((e, t) => e.persistenceId === t.persistenceId && e.sequenceNumber === t.sequenceNumber)
    }
```
But i have some questions about how to switch the projection to read the new way during the migration (step 2). When the projection reads the new table, which offset should be used as the starting point.

core/src/main/resources/reference.conf
```diff
@@ -134,6 +134,17 @@ jdbc-journal {
     }
 
     event_tag {
+      tableName = "event_tags"
```
patriknw, line 137: Do we have to change the table name? Is that because it's not possible to alter the primary key? Wouldn't it be difficult to keep the old tags when migrating?

I would like to see the migration instructions added to docs/src/main/paradox/migration.md

core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
```diff
@@ -23,20 +23,37 @@ class JournalQueries(
...
+      JournalTableC ++= events
+    }
+  }
+
+  private def writeEventsAndTags(sorted: Seq[(JournalAkkaSerializationRow, Set[String])])(
+      implicit ec: ExecutionContext): DBIOAction[Any, NoStream, Effect.Write] = {
+    val (events, tags) = sorted.unzip
+    // using revision by default.
+    val tagRows = sorted.map { case (eventRow, tags) =>
+      tags.map(t => TagRow(Option.empty, Option(eventRow.persistenceId), Option(eventRow.sequenceNumber), t))
```
patriknw, line 43: we typically write `None` instead of `Option.empty`, and `Some(...)` when it's clear that it will have a value

core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
```diff
@@ -23,20 +23,37 @@ class JournalQueries(
...
+    }
+    if (tagTableCfg.writeRedundant) {
       for {
         ids <- insertAndReturn ++= events
-        tagInserts = ids.zip(tags).flatMap { case (id, tags) => tags.map(tag => TagRow(id, tag)) }
+        tagInserts = ids.zip(tagRows).flatMap { case (id, tags) => tags.map(tag => tag.copy(Option(id))) }
```
patriknw, line 48: `tag.copy(eventId = Some(id))`

core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
```diff
@@ -23,20 +23,37 @@ class JournalQueries(
...
+      val events = sorted.map(_._1)
+      JournalTableC ++= events
+    }
+  }
+
+  private def writeEventsAndTags(sorted: Seq[(JournalAkkaSerializationRow, Set[String])])(
+      implicit ec: ExecutionContext): DBIOAction[Any, NoStream, Effect.Write] = {
+    val (events, tags) = sorted.unzip
+    // using revision by default.
+    val tagRows = sorted.map { case (eventRow, tags) =>
```
patriknw, line 42: I think it would be easier to understand if this is moved to the two `if` cases and then constructed with the eventId or not, instead of the additional map + copy

patriknw (commented):
looking good

Roiocam:
## proposal

After several months, I finally have the time to complete this merge. I've thoroughly reviewed the migration's changes, and here are my considerations.

Regarding the `event_tag` table, there are two essential components:

- The `ordering` used to sequence events in a specific stream.
- The `foreign key` used to load event payloads lazily.

In the previous approach, we used the `event_id` as the `foreign key`. However, this incurred some costs since we had to wait for all events to be inserted before retrieving their IDs. Moreover, the 'slick' plugin had performance issues with 'InsertAndReturn', which led to individual inserts instead of batch inserts.

On the other hand, we can employ the primary key of the `event_journal` table instead of the 'ordering' column for the foreign key. This modification removes the need for `InsertAndReturn`.

To ensure a rolling updates when shifting to the "new way," we propose a phased rollout with steps controlled by a configuration property:

1. Continue with the old approach.
2. Modify the table structure to allow for a redundant foreign key column, although only one will be used.
3. Write both old and new foreign keys, but only read the old foreign key for event lazy-loading (configurable through the configuration property).
4. Once the projection catches up with the newly written foreign key's position, migrate the table to the new primary and foreign keys. Alter the table structure to make the old foreign key column nullable.
5. Finally, the migration has completed, switch the projection to read the new foreign key and stop writing in the old manner (controlled by configuration properties).

### A real world example using MySQL.


### 1. add new column before migration

```sql
ALTER TABLE event_tag
    ADD PERSISTENCE_ID  VARCHAR(255),
    ADD SEQUENCE_NUMBER BIGINT;
```

### 2. change to redundant write and read via config

```config
jdbc-journal.tables.event_tag.redundant-write = true
jdbc-read-journal.tables.event_tag.redundant-read = true
```

### 3. waitting for projection catech, and then

```sql
-- drop old fk column
DELETE
FROM event_tag
WHERE PERSISTENCE_ID IS NULL
  AND SEQUENCE_NUMBER IS NULL;
-- drop old FK constraint
SELECT CONSTRAINT_NAME
INTO @fk_constraint_name
FROM INFORMATION_SCHEMA.REFERENTIAL_CONSTRAINTS
WHERE TABLE_NAME = 'event_tag';
SET @alter_query = CONCAT('ALTER TABLE event_tag DROP FOREIGN KEY ', @fk_constraint_name);
PREPARE stmt FROM @alter_query;
EXECUTE stmt;
DEALLOCATE PREPARE stmt;
-- drop old PK  constraint
ALTER TABLE event_tag
DROP PRIMARY KEY;
-- create new PK constraint for PK column.
ALTER TABLE event_tag
    ADD CONSTRAINT
        PRIMARY KEY (PERSISTENCE_ID, SEQUENCE_NUMBER, TAG);
-- create new FK constraint for PK column.
ALTER TABLE event_tag
    ADD CONSTRAINT fk_event_journal_on_pk
        FOREIGN KEY (PERSISTENCE_ID, SEQUENCE_NUMBER)
            REFERENCES event_journal (PERSISTENCE_ID, SEQUENCE_NUMBER)
            ON DELETE CASCADE;
-- alter the event_id to nullable, so we can skip the InsertAndReturn.
ALTER TABLE event_tag
    MODIFY COLUMN EVENT_ID BIGINT UNSIGNED NULL;
```

### 4. finally, rollback the redundant config


```config
jdbc-journal.tables.event_tag.redundant-write = false
jdbc-read-journal.tables.event_tag.redundant-read = false
```

core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
Roiocam: Thanks for the suggestions, i have been fix this on the newest commit.

core/src/main/resources/schema/postgres/postgres-create-schema.sql
```diff
@@ -23,11 +23,13 @@ CREATE UNIQUE INDEX event_journal_ordering_idx ON public.event_journal(ordering)
...
+    sequence_number BIGINT,
     tag VARCHAR(256),
-    PRIMARY KEY(event_id, tag),
+    PRIMARY KEY(persistence_id, sequence_number, tag),
     CONSTRAINT fk_event_journal
-      FOREIGN KEY(event_id)
-      REFERENCES event_journal(ordering)
+      FOREIGN KEY(persistence_id, sequence_number)
+      REFERENCES event_journal(persistence_id, sequence_number)
       ON DELETE CASCADE
```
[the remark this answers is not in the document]
Roiocam: I have confirmed this approach. It can first add the column and then, towards the end of the migration, finalize the changes to the table's primary key and foreign key.

core/src/test/scala/akka/persistence/jdbc/query/EventsByTagMigrationTest.scala
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    implicit val system: ActorSystem = ActorSystem("migrator-test", redundantConfig)
+    f(system)
+    system.terminate().futureValue
+  }
+
+  it should "migrate event tag to new way" in {
+    // 1. Mock legacy data on here, but actually using redundant write and read.
+    withRollingUpdateActorSystem { implicit system =>
```
Roiocam, line 123: During the verify step I commented these line here, and using the original approach to insert the same rows (at this time, the table didn't have any new column of journal table PK). By doing this, the outcome aligns more closely with real-world scenarios.

However, in order to avoid redundancy with the old schema, I'm use the new approach here to simulate the insertion.

I've validated this with databases other than H2, as I believe that when using H2, it implies that migration is not necessary.

patriknw (approved):
LGTM, thanks for great work on this

migrator/src/main/scala/akka/persistence/jdbc/migrator/JournalMigrator.scala
```diff
@@ -139,7 +139,14 @@ final case class JournalMigrator(profile: JdbcProfile)(implicit system: ActorSys
...
 
     val tagInserts =
-      newJournalQueries.TagTable ++= tags.map(tag => TagRow(journalSerializedRow.ordering, tag)).toSeq
+      newJournalQueries.TagTable ++= tags
+        .map(tag =>
+          TagRow(
+            Some(journalSerializedRow.ordering),
+            Some(journalSerializedRow.persistenceId),
+            Some(journalSerializedRow.sequenceNumber),
+            tag))
```
octonato, line 148: This migration tool is intended to migrate from the previous schema (without `event_tag` table) to the current one with the `event_tag` table. 

So in that sense, we could set the ordering to `None` or use the redundant flag to decide whether we want the order.

core/src/main/resources/schema/postgres/postgres-event-tag-migration.sql
```diff
@@ -0,0 +1,30 @@
+-- >>>>>>>>>>> before rolling updates.
+
+ALTER TABLE EVENT_TAG
+    ADD PERSISTENCE_ID  VARCHAR(255),
+    ADD SEQUENCE_NUMBER BIGINT;
+
+-- >>>>>>>>>>> after projection catch up.
```
octonato, line 7: It's not clear to me what do you mean by projection catch-up? Are you referring to the migration tool? The tool is intended for migration from previous versions. 

With the changes in this PR, we need to run another kind of migration.

Luckily, we can do this migration with plain SQL. For example, for Postgres, we can run the following:

```sql
UPDATE event_tag
  SET
    persistence_id = event_journal.persistence_id,
    sequence_number = event_journal.sequence_number
  FROM event_journal
  WHERE event_journal.ordering = event_tag.event_id;
```

After that, we can create the new foreign key and even drop the event_id.

core/src/main/resources/reference.conf
```diff
@@ -138,9 +138,17 @@ jdbc-journal {
...
         eventId = "event_id"
+        persistenceId = "persistence_id"
+        sequenceNumber = "sequence_number"
         tag = "tag"
       }
+
+      # For rolling updates the event_tag table migration.
+      # switch those to enabled new foregion key wirte and read.
+      redundant-write = false
+      redundant-read = false
```
octonato, line 151: I find it odd to call it 'redundant'. It looks more like a legacy schema.

octonato (commented):
Looking good. 

I think we can improve the migration instructions. We also need a page about it. 

For example, we need to ask users to create new columns and deploy a new version using `redundant-write`.

Then users still need to run an SQL script to backfill the PID and SeqNr. Something like: 

```scala
UPDATE event_tag
  SET
    persistence_id = event_journal.persistence_id,
    sequence_number = event_journal.sequence_number
  FROM event_journal
  WHERE event_journal.ordering = event_tag.event_id;
```

After that, deploy once more with `redundant-read` plus changes in the schema.

About `redundant` reads and writes, I find the name ambiguous. I made a comment in my comments below. 

I think we can even use one single config. For example, `legacy-tag-key`. By default set to true.

Users should add the new columns and update the plugin. 
The application will start to write in all three columns. 

After backfilling the two new columns (pid and seq_nr) with data from event_journal, the user deploys once again with `legacy-tag-key=false`. 
Now, the application won't write the `event_id` anymore and will always make the join using pid and seq_nr.

migrator/src/main/scala/akka/persistence/jdbc/migrator/JournalMigrator.scala
Roiocam: I have noticed that when verifying my tests, there doesn't seem to be much of a difference between them. Therefore, I did not make any changes.

core/src/main/resources/schema/postgres/postgres-event-tag-migration.sql
Roiocam: "after projection catch-up" refers to the process of waiting for Akka Projection to update the offset to the earliest new column write.

I hadn't considered the option of migrating data using SQL, which is great but I overlooked it. Including this SQL in the migration steps allows us to proceed without waiting for the Projection read "catch up".

I will update PR after fix integration test.

octonato: For clarification, there is no projection involved here. 

The tags are written when the events are persisted, not when consumed by a Projection. 

So basically, the journal stays what it is. Old tags will be written with ordering filled and pid and seq_nr will stay empty. New tags will have pid and seq_nr filled. Without explicit migration, nothing will happen.

Btw, Akka Projection is a separate project and may not be in use at all. You can use Akka Persistence with this plugin without even using Akka Projection.

patriknw:
Could you add the following to a file: `akka-persistence-jdbc/core/src/main/mima-filters/5.2.1.backwards.excludes/issue-710-tag-fk.excludes`

```
ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.persistence.jdbc.journal.dao.JournalTables#EventTags.eventId")
ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.persistence.jdbc.journal.dao.JournalTables#TagRow.eventId")
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.persistence.jdbc.journal.dao.JournalTables#TagRow.copy")
ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.persistence.jdbc.journal.dao.JournalTables#TagRow.copy$default$1")
ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.persistence.jdbc.journal.dao.JournalTables#TagRow.copy$default$2")
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.persistence.jdbc.journal.dao.JournalTables#TagRow.this")
ProblemFilters.exclude[MissingTypesProblem]("akka.persistence.jdbc.journal.dao.JournalTables$TagRow$")
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.persistence.jdbc.journal.dao.JournalTables#TagRow.apply")
ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.persistence.jdbc.journal.dao.JournalTables#TagRow.unapply")
```

patriknw:
There are also some test failures from CI if you can take a look at those?

Roiocam:
> Could you add the following to a file: `akka-persistence-jdbc/core/src/main/mima-filters/5.2.1.backwards.excludes/issue-710-tag-fk.excludes`

of course.

> There are also some test failures from CI if you can take a look at those?

Renato has some suggestions on this PR that sparked some thoughts in me: Perhaps we can simplify the rolling update steps by migrating SQL. 

I have some thoughts, but I am currently trapped by other issues and don't have the time for it at the moment. I will fix and verify them over the weekend.

Roiocam:
The migration guide will be add it on new pull request.

Roiocam:
@octonato could you please take a look agin? Thanks.

core/src/test/scala/akka/persistence/jdbc/query/EventsByTagMigrationTest.scala
```diff
@@ -0,0 +1,266 @@
...
+          tp.cancel()
+        }
+      }(system)
+    }
+
+    // Assume that the user could alter table for the addition of the new column manually, then we don't need to maintain
+    // the legacy table schema creation.
+    if (newDao) {
+      addNewColumn();
+      migrateLegacyRows();
```
Roiocam, line 195: using the SQL migration old rows, then we could rolling updates.

migrator/src/main/scala/akka/persistence/jdbc/migrator/JournalMigrator.scala
```diff
@@ -139,7 +139,14 @@ final case class JournalMigrator(profile: JdbcProfile)(implicit system: ActorSys
       .forceInsert(journalSerializedRow)
 
     val tagInserts =
-      newJournalQueries.TagTable ++= tags.map(tag => TagRow(journalSerializedRow.ordering, tag)).toSeq
+      newJournalQueries.TagTable ++= tags
+        .map(tag =>
+          TagRow(
+            Some(journalSerializedRow.ordering), // legacy tag key enabled by default.
```
octonato, line 145: This is fine, but we could also read the config key and then decide if we want to fill this column or not.

Ultimately, it will be nicer if users can even not have this column in their table.

core/src/main/resources/schema/h2/h2-create-schema.sql
```diff
@@ -18,12 +18,14 @@ CREATE TABLE IF NOT EXISTS "event_journal" (
 CREATE UNIQUE INDEX "event_journal_ordering_idx" on "event_journal" ("ordering");
 
 CREATE TABLE IF NOT EXISTS "event_tag" (
-    "event_id" BIGINT NOT NULL,
+    "event_id" BIGINT,
```
octonato, line 21: It would be better if new users just don't have this column.

I know, legacy mode is enabled by default and if you just start today with the plugin you will be using it in legacy mode without even noticing it. 

I don't have a solution. I'm just mentioning it so we can think together about alternatives. 

In the worst case scenario, we leave it as is and people will have a column that is never used.

octonato: Probably, we need to go with that and on a next release we simply drop this column and we remove the legacy mode flag. 

Then users will first need to move to the previous version, run the migrate and then move to the next version. After all that they will be able to drop the `event_id` column.

octonato (approved):
LGTM

I left a few comments but more about things for later we should consider after we cut a release with this change.

core/src/main/resources/schema/mysql/mysql-event-tag-migration.sql
```diff
@@ -0,0 +1,34 @@
+-- add new column
+ALTER TABLE event_tag
+    ADD persistence_id  VARCHAR(255),
+    ADD sequence_number BIGINT;
+-- migrate rows
```
octonato, line 5: I just realised that the script needs to be run in two parts. First add the columns, redeploy with new version, then run the rest of the script and redeploy once more with legacy-mode set to false. See my comment on the migration guide PR: https://github.com/akka/akka-persistence-jdbc/pull/781#pullrequestreview-1698857285

octonato (commented):
@Roiocam, thanks for all the efforts and patience. I have one more comment that I think we need to address before merging. 

We are almost there.

PR #732, opened 2023-04-13
chore: compatible with Windows OS, don't format LF to CRLF

Roiocam:
No References


johanandren:
Shouldn't this be handled by your git settings rather? Setting `core.autocrlf=true` turns LF to CRLF on checkout and back on commit.

Issue #673, opened 2022-09-05
failure: Some eventsByTag tests fail with Oracle and legacy.ByteArrayReadJournalDao


patriknw:
Started when logical deletion was removed in https://github.com/akka/akka-persistence-jdbc/pull/569, but that probably didn't work with logicalDelete=false before that either. It showed up with that change because the default for logical delete was changed.

Roiocam:
I believe the reason for this issue is because Oracle does not support the boolean data type. Therefore, after PR #569, the functionality stopped working. 

It's frustrating because this problem has recurred in PR #731 and has caused me some difficulty.

![截屏2023-09-05 13 38 39](https://github.com/akka/akka-persistence-jdbc/assets/26020358/5fa2ab55-7d04-4f58-9d40-afd4a4a261c8)

PR #764, opened 2023-09-06
fix: Avoid using boolean types under Oracle's data type limitations #673

Roiocam:
References #673

This PR has been fix something like `EventsByTagTest`, but i got the ERROR on `JournalSequenceActorTest`.

I think it was my environment complaining database too slow, waiting for investigation.

### DIFF
--- a/core/src/main/scala/akka/persistence/jdbc/query/dao/legacy/ByteArrayReadJournalDao.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/query/dao/legacy/ByteArrayReadJournalDao.scala
@@ -122,7 +122,7 @@ trait OracleReadJournalDao extends ReadJournalDao {
     if (isOracleDriver(profile)) {
       val theOffset = Math.max(0, offset)
       val theTag = s"%$tag%"
-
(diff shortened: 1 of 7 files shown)


Roiocam:
@patriknw These CI checks have passed. Does this mean that this PR has successfully solved #673?

core/src/main/scala/akka/persistence/jdbc/query/dao/legacy/ByteArrayReadJournalDao.scala
```diff
@@ -122,7 +122,7 @@ trait OracleReadJournalDao extends ReadJournalDao {
     if (isOracleDriver(profile)) {
       val theOffset = Math.max(0, offset)
       val theTag = s"%$tag%"
-
+      // fixes https://github.com/akka/akka-persistence-jdbc/issues/673, oracle do not support boolean type.
```
johanandren, line 125: No need to leave a comment around, the commit history will tell the story
```suggestion
```

johanandren (commented):
One small thing then I think this looks good to merge

johanandren (approved):
LGTM, thanks!

PR #765, opened 2023-09-14
perf: Reducing memory cost while replaying

Roiocam:
No References.

#### Motivation:

Reducing memory cost on replaying query stream

#### Result

Additional bonus: slight performance improvement.

```bash
jmh:run -i 3 -wi 3 -f 1 .JdbcAsyncJournalBenchmark

[info] Benchmark                                     Mode  Cnt       Score       Error  Units
[info] JdbcAsyncJournalBenchmark.originalAsync      thrpt    3  590394.058 ± 38433.920  ops/s
[info] JdbcAsyncJournalBenchmark.streamWithCollect  thrpt    3  594978.499 ±  2034.776  ops/s

[info] Benchmark                                     Mode  Cnt       Score       Error  Units
[info] JdbcAsyncJournalBenchmark.originalAsync      thrpt    3  584438.653 ± 35769.421  ops/s
[info] JdbcAsyncJournalBenchmark.streamWithCollect  thrpt    3  587042.327 ± 15553.554  ops/s
```

JMH for this change: https://gist.github.com/Roiocam/aabbf404aa2e1a4e967967e450b0def0

### DIFF
--- a/core/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
@@ -118,7 +118,7 @@ class JdbcAsyncWriteJournal(config: Config) extends AsyncWriteJournal {
     journalDao
       .messagesWithBatch(persistenceId, fromSequenceNr, toSequenceNr, journalConfig.daoConfig.replayBatchSize, None)
       .take(max)
-      .mapAsync(1)(reprAndOrdNr => Future.fromTry(reprAndOrdNr))
+      .collect { case Success(reprAndOrdNr) => reprAndOrdNr }
       .runForeach { case (repr, _) =>
         recoveryCallback(repr)
       }
--- a/core/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
@@ -118,9 +118,10 @@ class JdbcAsyncWriteJournal(config: Config) extends AsyncWriteJournal {
     journalDao
       .messagesWithBatch(persistenceId, fromSequenceNr, toSequenceNr, journalConfig.daoConfig.replayBatchSize, None)
       .take(max)
-      .collect { case Success(reprAndOrdNr) => reprAndOrdNr }
-      .runForeach { case (repr, _) =>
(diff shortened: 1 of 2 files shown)


core/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
```diff
@@ -118,7 +118,7 @@ class JdbcAsyncWriteJournal(config: Config) extends AsyncWriteJournal {
     journalDao
       .messagesWithBatch(persistenceId, fromSequenceNr, toSequenceNr, journalConfig.daoConfig.replayBatchSize, None)
       .take(max)
-      .mapAsync(1)(reprAndOrdNr => Future.fromTry(reprAndOrdNr))
+      .collect { case Success(reprAndOrdNr) => reprAndOrdNr }
```
johanandren, line 121: This has different behaviour than the original logic though, that would fail the stream on a Failure while this filters it out. Better would be to skip this operator/step completely and do the match in `runForEach` and throw for Failure there:

```scala
      .runForeach { 
        case Success((repr, _)) =>
          recoveryCallback(repr)
        case Failure(ex) => throw ex
      }
```

johanandren (changes requested):

core/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala
Roiocam: More effiency, Yay!

```
[info] Benchmark                                     Mode  Cnt       Score        Error  Units
[info] JdbcAsyncJournalBenchmark.originalAsync      thrpt    3  581405.864 ± 217516.676  ops/s
[info] JdbcAsyncJournalBenchmark.streamWithCollect  thrpt    3  594420.619 ±  29680.962  ops/s

[info] Benchmark                                     Mode  Cnt       Score       Error  Units
[info] JdbcAsyncJournalBenchmark.originalAsync      thrpt    3  577291.504 ± 51386.930  ops/s
[info] JdbcAsyncJournalBenchmark.streamWithCollect  thrpt    3  605493.214 ±  4152.262  ops/s
```

Roiocam: Push my commit again, because i havn't permission to apply this request change,

johanandren (approved):
LGTM, thanks!

PR #781, opened 2023-10-26
docs: event tag migration guide

Roiocam:
References #731

### DIFF
--- a/docs/src/main/paradox/migration.md
+++ b/docs/src/main/paradox/migration.md
@@ -6,7 +6,7 @@ Release `5.4.0` change the schema of `event_tag` table.
 
 The previous version was using an auto-increment column as a primary key and foreign key on the `event_tag` table. As a result, the insert of multiple events in batch was not performant.
 
-while in `5.4.0` the primary key and foreign key of `event_tag` table has been replaced with primary key of `event_journal` table. In order to migrate to the new schema, we made a [**migration script**](https://github.com/akka/akka-persistence-jdbc/tree/master/core/src/main/resources/schema) capable of the column creation, rows and constraint migration.
+While in `5.4.0`, the primary key and foreign key on the `event_tag` table have been replaced with a primary key from the `event_journal` table. In order to migrate to the new schema, we made a [**migration script**](https://github.com/akka/akka-persistence-jdbc/tree/master/core/src/main/resources/schema) which is capable of creating the new column, migrate the rows and add the new constraints.
 
 The new "tag key" disable by default, If you want to use it, apply migration script and enable property via configuration:
 
--- a/docs/src/main/paradox/migration.md
+++ b/docs/src/main/paradox/migration.md
@@ -4,7 +4,7 @@
 
 Release `5.4.0` change the schema of `event_tag` table.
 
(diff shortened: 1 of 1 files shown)


octonato (commented):
I left some rewording suggestions.

I know realise that the way we are creating the migration script wouldn't allow for proper rolling upgrades. 

To have it properly done, a user must first run a script that only creates the new columns. 

Then they need to redeploy their application in legacy-mode. The application will start to write on all three columns. But the join will be done using the `event_id`. 

After that, they should run the second part of the script that will migrate the rows and adapt the constraints. Once this is done, they can redeploy their application by disabling the legacy-mode.

docs/src/main/paradox/migration.md
```diff
@@ -1,5 +1,34 @@
 # Migration
 
+## Migrating to version 5.4.0
+
+Release `5.4.0` change the schema of `event_tag` table.
+
+The previous version was using auto increment column as a primary key and foreign key of `event_tag` table, this has resulted in we can no longer do a batch insert due to the insert returning the ordering.
```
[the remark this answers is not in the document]
Roiocam: make sense.

Roiocam:
@octonato I have update the guide, Explicitly point the mutiple-phase rollout/deploy, and the two configuration style.

The script has update this too: https://github.com/akka/akka-persistence-jdbc/pull/731/commits/8587ba03b64e214a8352b72d0a2446136bce56de

Can you take a look? Thanks.

docs/src/main/paradox/migration.md
```diff
@@ -1,5 +1,35 @@
...
 
+## Migrating to version 5.4.0
+
+Release `5.4.0` change the schema of `event_tag` table.
+
+The previous version was using an auto-increment column as a primary key and foreign key on the `event_tag` table. As a result, the insert of multiple events in batch was not performant.
+
+While in `5.4.0`, the primary key and foreign key on the `event_tag` table have been replaced with a primary key from the `event_journal` table. In order to migrate to the new schema, we made a [**migration script**](https://github.com/akka/akka-persistence-jdbc/tree/master/core/src/main/resources/schema) which is capable of creating the new column, migrate the rows and add the new constraints.
+
+By default, the plugin will behave as in previous version. If you want to use the new `event_tag` keys, you need to multiple-phase rollout:
```
octonato, line 11:
```suggestion
By default, the plugin will behave as in previous version. If you want to use the new `event_tag` keys, you need to run a multiple-phase rollout:
```

docs/src/main/paradox/migration.md
```diff
@@ -1,5 +1,35 @@
...
+
+1. apply the first part of the migration script and then redeploy your application on default settings.
```
octonato, line 13:
```suggestion
1. apply the first part of the migration script and then redeploy your application with the default settings.
```

octonato (approved):
Thanks @Roiocam 

Looking good. Left a few minor change requests. 
I think we can merge it after that.

Roiocam:
> Thanks @Roiocam
> 
> Looking good. Left a few minor change requests. I think we can merge it after that.

Thanks for the feedback.

PR #775, opened 2023-10-16
Scala 3 support

jtjeferreira:
Adds scala 3 support

PR #804, opened 2024-03-08
Version bumps - especially Slick 3.5.0

JustinPihony:


johanandren:
Ok, obviously not possible to bump without lots of other changes, we already have a contribution in progress that should deal with that in #775

PR #805, opened 2024-03-08
build: Incorrect flags for Scala 3 scalac blocking publish

johanandren:
Something we missed in #775 because only triggered by publish.


ennru (approved):
LGTM.

Issue #816, opened 2024-03-22
Problems with SLF4J


johanandren:
In 5.4.0 we bumped Slick to 3.5.0, which transitively pulls in slf4-api 2.2. Unfortunately slf4j had a binary break between 1.7 which is what is supported by Akka and 2.2, so this will cause grief for users. For example it may work without any linker/classdef not found errors but not find the configured logger.

Typical symptoms of this are having logback on classpath but seeing this on startup:
```
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
SLF4J: Class path contains SLF4J bindings targeting slf4j-api versions 1.7.x or earlier.
SLF4J: Ignoring binding found at [jar:file:/C:/Users/user/.m2/repository/ch/qos/logback/logback-classic/1.2.11/logback-classic-1.2.11.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Ignoring binding found at [jar:file:/C:/Users/user/.m2/repository/org/apache/logging/log4j/log4j-slf4j-impl/2.19.0/log4j-slf4j-impl-2.19.0.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See https://www.slf4j.org/codes.html#ignoredBindings for an explanation.
```

Would be good if we could somehow override/exclude the 2.2 dependency and force sl4j-api 1.7

sebastian-alfers:
I used the approach from the PR in some sample project but that did not help, I had to use `dependencyOverrides` as well.

johanandren:
Do you have some more details, we have hit this a couple times in Akka projects but exclude from jdbc (or Slick) has been enough. For example: https://github.com/akka/akka-persistence-r2dbc/pull/541/files#diff-ad2642dc77b3679e4ad2c7d3b2f59a2dcf48c1bf5b93a558a4fa44828315a34cL75

Would be good to know in what scenario that doesn't work.

johanandren:
Sebastian shared a reproducer with me and the problem is when pulling in both akka-persistence-jdbc, slick itself, and slick-hickaricp. Each of those would need a `exclude("org.slf4j", "slf4j-api")` so it is easier to override that dependency with only one `dependencyOverrides += "org.slf4j" % "slf4j-api" % "1.7.36"` entry in the build.

Probably messy on the maven side as well.

PR #817, opened 2024-03-22
chore: Pin slf4j-api to 1.7 to avoid downstream problems with Akka

johanandren:
Refs #816

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -21,7 +21,9 @@ object Dependencies {
 
   val Libraries: Seq[ModuleID] = Seq(
     "com.typesafe.akka" %% "akka-persistence-query" % AkkaVersion,
-    "com.typesafe.slick" %% "slick" % SlickVersion,
+    // Slick 3.5 pulls in slf4j-api 2.2 which doesn't work with Akka
+    ("com.typesafe.slick" %% "slick" % SlickVersion).exclude("org.slf4j", "slf4j-api"),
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -21,7 +21,9 @@ object Dependencies {
 
   val Libraries: Seq[ModuleID] = Seq(
     "com.typesafe.akka" %% "akka-persistence-query" % AkkaVersion,
-    "com.typesafe.slick" %% "slick" % SlickVersion,
+    // Slick 3.5 pulls in slf4j-api 2.2 which doesn't work with Akka
+    ("com.typesafe.slick" %% "slick" % SlickVersion).exclude("org.slf4j", "slf4j-api"),
+    "org.slf4j" % "slf4j-api" % "1.7.36",
```
patriknw, line 26: do we need to add the explicit version here? with the exclude in place, I thought it would use what we have defined in upstream Akka dependencies

would be easy to forget this one once we bump slf4j in Akka

johanandren: I was thinking that since we remove the transitive one, we need to make sure there is another one pulled in or else you could end up with a classdef not found in some scenario, users will most likely have the akka-slf4j in there but maybe not always?

patriknw: ok, true, if not using Akka Typed
I'll add a note in the slf4j update ticket

patriknw (approved):
LGTM

PR #819, opened 2024-03-25
build: Fold integration tests into one build matrix

johanandren:
Also: add nightly build matrix on JDK 21


ennru (approved):
Simpler than anticipated.
LGTM.

PR #826, opened 2024-04-19
perf: improve deletion, avoid multiple rows updating.

Roiocam:
Nothing related it. Just improve the performance of deletion by avoiding multiple row updates.


johanandren:
Now fixed by #847 (sorry, we missed that this PR existed)

PR #831, opened 2024-04-26
fix not working JournalDaoStreamMessagesMemoryTest

Roiocam:
JournalDaoStreamMessagesMemoryTest didn't work in the newDao because it uses legacyDao, we could create dao instance via reflection to make it works.


Roiocam:
@johanandren @octonato would you like to take a look?

octonato:
Was this test not being run? How it was passing before?

johanandren (approved):
LGTM, I verified in test runs that it is executing and pass (takes 42s on oracle though 😱)

johanandren:
I see I missed one thing though, we lost the `class H2JournalDaoStreamMessagesMemoryTest`, I'll follow up and re-introduce

PR #833, opened 2024-05-07
chore: reintroduce H2JournalDaoStreamMessagesMemoryTest

johanandren:
Accidentally removed in #831


johanandren:
Ah, I see now how that was not unintentional, H2 will use a lot of the heap so that test becomes close to useless.

PR #834, opened 2024-05-08
build: Akka 2.9.3 bump, Akka Paradox Theme bump, 5.4.1 license change date

johanandren:


ennru (approved):
LGTM.

core/src/main/scala/akka/persistence/jdbc/query/scaladsl/JdbcReadJournal.scala
```diff
@@ -108,8 +106,9 @@ class JdbcReadJournal(config: Config, configPath: String)(implicit val system: E
    */
   override def persistenceIds(): Source[String, NotUsed] =
     Source
-      .repeat(0)
-      .flatMapConcat(_ => delaySource.flatMapConcat(_ => currentPersistenceIds()))
+      .single(0)
+      .concat(Source.tick(readJournalConfig.refreshInterval, readJournalConfig.refreshInterval, 0))
```
patriknw, line 110: This works, but why do we need single+concat? Wouldn't an initial delay of 0 be enough? Alright, initial delay isn't allowed to be 0? Wonder if we shouldn't allow initial 0 and also allow 0 for scheduleOnce?

Issue #824, opened 2024-04-16
Durable State broken after upgrading to slick 3.5.x


sebastian-alfers:
When persisting a Durable State with slick 53.5.1, the following two errors show up:

```
ava.lang.NoClassDefFoundError: scala/reflect/api/Trees$TreeApi
	at slick.lifted.AnyOptionExtensionMethods$.getOrElse$extension(ExtensionMethods.scala:219)
	at akka.persistence.jdbc.state.DurableStateQueries.<init>(DurableStateQueries.scala:103)
	at akka.persistence.jdbc.state.javadsl.JdbcDurableStateStore.<init>(JdbcDurableStateStore.scala:40)
	at akka.persistence.jdbc.state.JdbcDurableStateStoreProvider.javadslDurableStateStore(JdbcDurableStateStoreProvider.scala:40)
	at akka.persistence.state.DurableStateStoreRegistry$$anon$1.javaDsl(DurableStateStoreRegistry.scala:41)
	at akka.persistence.state.DurableStateStoreRegistry$$anon$1.javaDsl(DurableStateStoreRegistry.scala:39)
	at akka.persistence.PersistencePlugin$$anon$1.createExtension(PersistencePlugin.scala:61)
	at akka.persistence.PersistencePlugin$$anon$1.createExtension(PersistencePlugin.scala:58)
	at akka.actor.ActorSystemImpl.registerExtension(ActorSystem.scala:1201)
	at akka.actor.ExtensionId.apply(Extension.scala:78)
	at akka.actor.ExtensionId.apply$(Extension.scala:77)
	at akka.persistence.PersistencePlugin$$anon$1.apply(PersistencePlugin.scala:58)
	at akka.persistence.PersistencePlugin.pluginFor(PersistencePlugin.scala:56)
	at akka.persistence.state.DurableStateStoreRegistry.durableStateStoreFor(DurableStateStoreRegistry.scala:81)
	at akka.persistence.typed.state.internal.BehaviorSetup.<init>(BehaviorSetup.scala:55)
	at akka.persistence.typed.state.internal.DurableStateBehaviorImpl.$anonfun$apply$2(DurableStateBehaviorImpl.scala:115)
	at akka.actor.typed.internal.BehaviorImpl$DeferredBehavior$$anon$1.apply(BehaviorImpl.scala:119)
	at akka.actor.typed.Behavior$.start(Behavior.scala:176)
[...]

PR #835, opened 2024-05-08
fix: Workaround slick dependency issue #824

johanandren:
Direct dependency on scala-reflect sorts it until upstream is fixed

References #824


ennru (approved):
LGTM.

PR #847, opened 2024-07-31
fix: delete all but last

octonato:
First delete all event but the last and then mark the last as delete.

### DIFF
--- a/core/src/main/scala/akka/persistence/jdbc/journal/dao/DefaultJournalDao.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/journal/dao/DefaultJournalDao.scala
@@ -50,8 +50,7 @@ class DefaultJournalDao(
 
   override def delete(persistenceId: String, maxSequenceNr: Long): Future[Unit] = {
     val actions: DBIOAction[Unit, NoStream, Effect.Write with Effect.Read] = for {
-      highestMarkedSequenceNrOpt <- highestMarkedSequenceNr(persistenceId)
-      highestMarkedSequenceNr = highestMarkedSequenceNrOpt.getOrElse(0L)
+      highestMarkedSequenceNr <- highestSequenceNrAction(persistenceId)
       _ <- queries.delete(persistenceId, highestMarkedSequenceNr - 1)
       _ <- queries.markAsDeleted(persistenceId, highestMarkedSequenceNr)
     } yield ()
@@ -64,6 +63,9 @@ class DefaultJournalDao(
       maybeHighestSeqNo <- db.run(queries.highestSequenceNrForPersistenceId(persistenceId).result)
     } yield maybeHighestSeqNo.getOrElse(0L)
   }
+
+  private def highestSequenceNrAction(persistenceId: String) =
+    queries.highestSequenceNrForPersistenceId(persistenceId).result.map(_.getOrElse(0))
 
   private def highestMarkedSequenceNr(persistenceId: String) =
     queries.highestMarkedSequenceNrForPersistenceId(persistenceId).result
--- a/core/src/main/scala/akka/persistence/jdbc/journal/dao/DefaultJournalDao.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/journal/dao/DefaultJournalDao.scala
@@ -58,11 +58,8 @@ class DefaultJournalDao(
     db.run(actions.transactionally)
   }
 
-  override def highestSequenceNr(persistenceId: String, fromSequenceNr: Long): Future[Long] = {
-    for {
-      maybeHighestSeqNo <- db.run(queries.highestSequenceNrForPersistenceId(persistenceId).result)
-    } yield maybeHighestSeqNo.getOrElse(0L)
-  }
+  override def highestSequenceNr(persistenceId: String, fromSequenceNr: Long): Future[Long] =
+    db.run(highestSequenceNrAction(persistenceId))
 
   private def highestSequenceNrAction(persistenceId: String): DBIOAction[Long, NoStream, Effect.Read] =
     queries.highestSequenceNrForPersistenceId(persistenceId).result.map(_.getOrElse(0))
--- a/core/src/main/scala/akka/persistence/jdbc/journal/dao/DefaultJournalDao.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/journal/dao/DefaultJournalDao.scala
@@ -64,7 +64,7 @@ class DefaultJournalDao(
     } yield maybeHighestSeqNo.getOrElse(0L)
   }
 
-  private def highestSequenceNrAction(persistenceId: String) =
+  private def highestSequenceNrAction(persistenceId: String): DBIOAction[Long, NoStream, Effect.Read] =
     queries.highestSequenceNrForPersistenceId(persistenceId).result.map(_.getOrElse(0))
 
   private def highestMarkedSequenceNr(persistenceId: String) =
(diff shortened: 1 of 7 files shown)


core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
```diff
@@ -69,10 +69,10 @@ class JournalQueries(
     JournalTable.filter(_.persistenceId === persistenceId).filter(_.sequenceNumber <= toSequenceNr).delete
   }
 
-  def markJournalMessagesAsDeleted(persistenceId: String, maxSequenceNr: Long) =
+  def markAsDeleted(persistenceId: String, highestMarkedSequenceNr: Long) =
     JournalTable
       .filter(_.persistenceId === persistenceId)
-      .filter(_.sequenceNumber <= maxSequenceNr)
+      .filter(_.sequenceNumber === highestMarkedSequenceNr)
```
octonato, line 75: being more explicitly and only marking a single message as delete. 

Only used when deleted events though and what we want is to delete the latest.

octonato: Hmm, this will probably not work as expected. Looking further in the code.

core/src/main/scala/akka/persistence/jdbc/journal/dao/DefaultJournalDao.scala
```diff
@@ -64,6 +64,9 @@ class DefaultJournalDao(
     } yield maybeHighestSeqNo.getOrElse(0L)
   }
 
+  private def highestSequenceNrAction(persistenceId: String) =
+    queries.highestSequenceNrForPersistenceId(persistenceId).result.map(_.getOrElse(0))
+
```
octonato, line 69: Added this extra query that return the highest seqNr (non-deleted).

core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
```diff
@@ -69,14 +69,22 @@ class JournalQueries(
...
-  def markAsDeleted(persistenceId: String, highestMarkedSequenceNr: Long) =
+  private[akka] def markAsDeleted(persistenceId: String, highestMarkedSequenceNr: Long) =
     JournalTable
       .filter(_.persistenceId === persistenceId)
       .filter(_.sequenceNumber === highestMarkedSequenceNr)
       .filter(_.deleted === false)
       .map(_.deleted)
       .update(true)
 
+  def markJournalMessagesAsDeleted(persistenceId: String, maxSequenceNr: Long) =
```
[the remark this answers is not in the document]
johanandren: Mark as deprecated?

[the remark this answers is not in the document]
octonato: With what message? We are not offering an alternative to it.

If we deprecate it and someone is using, it more like: we will remove it and you have to give up from whatever you are doing with this method because there is no alternative. 

But I might be overthinking it. This is mostly internal API.

core/src/main/scala/akka/persistence/jdbc/journal/dao/DefaultJournalDao.scala
```diff
@@ -48,21 +52,34 @@ class DefaultJournalDao(
...
+            // if so, we delete up to the before last and
+            // mark the last as logically deleted preserving highestSeqNr
+            queries
+              .delete(persistenceId, highestSeqNr - 1)
+              .flatMap(_ => queries.markAsDeleted(persistenceId, highestSeqNr))
+          case _ =>
+            // if not, we delete up to the requested seqNr
+            queries.delete(persistenceId, toSequenceNr)
+        }
+        .map(_ => ())
```
octonato, line 73: Tests for h2 are passing now locally. Should pass for the the other databases as well. All failures were comping from the akka journal specs.

core/src/main/scala/akka/persistence/jdbc/journal/dao/DefaultJournalDao.scala
```diff
@@ -48,21 +52,34 @@ class DefaultJournalDao(
   val queries =
     new JournalQueries(profile, journalConfig.eventJournalTableConfiguration, journalConfig.eventTagTableConfiguration)
 
-  override def delete(persistenceId: String, maxSequenceNr: Long): Future[Unit] = {
-    val actions: DBIOAction[Unit, NoStream, Effect.Write with Effect.Read] = for {
-      _ <- queries.markJournalMessagesAsDeleted(persistenceId, maxSequenceNr)
-      highestMarkedSequenceNr <- highestMarkedSequenceNr(persistenceId)
-      _ <- queries.delete(persistenceId, highestMarkedSequenceNr.getOrElse(0L) - 1)
```
octonato, line 55: This was indeed not efficient and there was also a little bug here. 

For a journal with events: 1, 2, 3, 4, 5

Deleting up to 3, would first:
logically delete: 1, 2, 3
collect the highest logically deleted nr, thus 3
delete everything until 3, so delete 1 and 2.

The journal remains with: 3, 4, 5 while 3 is invisible because logically deleted. 

There is no reason to keep 3, because it's not the highest anyway.

core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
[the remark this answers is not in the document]
octonato: I added this:

`@deprecated(message = "Intended to be internal API", since = "5.4.2")`

core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
```diff
@@ -69,6 +69,15 @@ class JournalQueries(
     JournalTable.filter(_.persistenceId === persistenceId).filter(_.sequenceNumber <= toSequenceNr).delete
   }
 
+  private[akka] def markAsDeleted(persistenceId: String, seqNr: Long) =
+    JournalTable
+      .filter(_.persistenceId === persistenceId)
+      .filter(_.sequenceNumber === seqNr)
+      .filter(_.deleted === false)
+      .map(_.deleted)
+      .update(true)
```
johanandren, line 78: My Slick knowledge is rusty, does the `.map(_.deleted)` somehow change the boolean value of the field `deleted` or what does this do?

core/src/main/scala/akka/persistence/jdbc/journal/dao/DefaultJournalDao.scala
```diff
@@ -48,21 +52,34 @@ class DefaultJournalDao(
...
-    } yield ()
+  override def delete(persistenceId: String, toSequenceNr: Long): Future[Unit] = {
+
+    // note: the passed toSequenceNr will be Long.MaxValue when doing a 'full' journal clean-up
+    // see JournalSpec's test: 'not reset highestSequenceNr after journal cleanup'
```
johanandren, line 58: Huh, surprising, I'd expect you often do delete(Long.Max) to drop all events for an entity (and expect to not break projections if re-using the pid) but ok, if the tests say so I guess I'm wrong.

johanandren (approved):
LGTM but two questions

octonato:
We are having issues with mysql tests. Container not starting. 3 days ago it passed. Might be a intermittent issue.

core/src/main/scala/akka/persistence/jdbc/journal/dao/JournalQueries.scala
johanandren: Thanks for clarifying

PR #849, opened 2024-08-28
fix: deserialization failure in durable state should result in failed future

leviramsey:
The present implementation of the durable state store treats a deserialization failure as "successful but absent" (a `Future` containing `None`)

In contrast:
* this plugin's journal and snapshot stores will fail the future (see [journal](https://github.com/akka/akka-persistence-jdbc/blob/55f38148ff5782a0d993b8f51b810f163160587e/core/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala#L124) and [snapshot](https://github.com/akka/akka-persistence-jdbc/blob/55f38148ff5782a0d993b8f51b810f163160587e/core/src/main/scala/akka/persistence/jdbc/snapshot/dao/DefaultSnapshotDao.scala#L72)
* [the R2DBC plugin's durable state store will fail the future](https://github.com/akka/akka-persistence-r2dbc/blob/93246b520a9efbdfd5d6b5205f80ab99fd9108f6/core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/R2dbcDurableStateStore.scala#L123)

With this change, the behavior now matches the others: a failed try from the serializer will be a failed future.

### DIFF
--- a/core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/state/scaladsl/JdbcDurableStateStore.scala
@@ -59,15 +59,14 @@ class JdbcDurableStateStore[A](
     s"akka-persistence-jdbc-durable-state-sequence-actor")
 
   def getObject(persistenceId: String): Future[GetObjectResult[A]] = {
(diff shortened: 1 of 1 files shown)


johanandren (approved):
LGTM

PR #908, opened 2025-04-30
fix: link validation for missing slick docsite

octonato:
Kinds of workaround for link validation. 

Slick doc-site for 3.5.2 is missing and we are not ready to bump to 3.6.0 (see https://github.com/akka/akka-persistence-jdbc/pull/902#issuecomment-2788446422)

So for now, I think the best we can do it to point to https://scala-slick.org/doc/stable/, although it might give the impression that we are on the latest version.

### DIFF
--- a/scripts/link-validator.conf
+++ b/scripts/link-validator.conf
@@ -14,6 +14,12 @@ site-link-validator {
       prefix = "https://doc.akka.io/api/akka-persistence-jdbc/snapshot/"
       replace = "/api/akka-persistence-jdbc/snapshot/"
     }
+    {
+      # slick 3.5.2 doc site is removed, next version is 3.6.0
+      # we keep this exclusion as long as we stay with 3.5.2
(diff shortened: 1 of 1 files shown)


octonato:
Ok, the reason why there is no docsite for 3.5.2 is that they consider it broken. 
https://github.com/slick/slick/issues/3030 

They accidentely published it while they dropped support for jdk8. 
It's not an issue for us it seems, except for the missing link. 

Maybe we can replace it with `3.5.1` instead of `stable`?

johanandren:
Sounds like we should roll back to slick 3.5.1?

octonato:
Did we release it with 3.5.2 already? If not, rollback might be better indeed.

octonato:
yes, it's released. But it might be find to revert for the next one.
I will send another PR and we can then decide.

johanandren:
I merged the other one, I think it is fine to revert since the Slick 3.5.2 seems to have been an accident and not an intended release.

PR #953, opened 2026-04-15
bump: Slick 3.6.1

sebastian-alfers:
To align with https://github.com/akka/akka-projection/pull/1429


johanandren (approved):
LGTM, let's merge this to keep up with slick, even if there are some bin comp issues with the bump.

PR #954, opened 2026-04-28
fix: quote identifiers in DurableState raw SQL paths

JustinPihony:
## Summary

`DurableStateQueries` builds INSERT/UPDATE statements with Slick's `#$` raw interpolation, which injects the table and column names **unquoted**, while the SELECT path (typed queries) quotes them via the profile. On H2 in its default mode unquoted identifiers are uppercased, so writes against the lowercase quoted schema (`schema/h2/h2-create-schema.sql`) fail:

```
Table "DURABLE_STATE" not found (candidates are: "durable_state")
```

The journal/snapshot plugins are unaffected — they use Slick's typed query API exclusively.

The existing H2 test configs hid this by passing `DATABASE_TO_UPPER=false` on the JDBC URL, which disables H2's default uppercasing.

## Fix

Quote `tableAndSchema` and every injected column name through `profile.quoteIdentifier(...)` in [`DurableStateQueries.scala`](core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala). The profile produces `"…"` for H2/Postgres/Oracle/SqlServer and backticks for MySQL — matching the quoting Slick already applies on the typed-query path and the schema DDL.

## Regression coverage

Adds `H2DefaultModeDurableStateSpec` running against a new `h2-default-mode-application.conf` that omits `DATABASE_TO_UPPER=false`. Without this fix the new spec reproduces the bug (14 failures with the `Table "DURABLE_STATE" not found` error); with the fix all 18 tests pass.

The shared `h2-application.conf` is intentionally left as-is — `H2ScalaEventsByTagMigrationTest` constructs unquoted DDL on purpose for the migration scenario and relies on the workaround.

## Test plan

- [x] `core/testOnly akka.persistence.jdbc.state.scaladsl.H2DurableStateSpec` (existing, passes)
- [x] `core/testOnly akka.persistence.jdbc.state.scaladsl.H2DefaultModeDurableStateSpec` (new regression spec, passes — and fails as expected when fix is reverted)
- [x] `core/testOnly akka.persistence.jdbc.query.H2ScalaEventsByTagMigrationTest` (existing, still passes)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
+++ b/core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
@@ -23,8 +23,21 @@ import akka.persistence.jdbc.config.DurableStateTableConfiguration
     extends DurableStateTables {
   import profile.api._
 
-  lazy val tableAndSchema = durableStateTableCfg.schemaName.fold(durableStateTableCfg.tableName)(schema =>
-    s"$schema.${durableStateTableCfg.tableName}")
+  // Identifiers must be quoted via the profile so the raw-SQL INSERT/UPDATE paths use the
+  // same quoting as Slick's typed-query SELECT path. Without this, e.g. H2 in default mode
+  // uppercases unquoted identifiers, which doesn't match the lowercase quoted identifiers
+  // used by the schema and Slick's typed queries.
+  lazy val tableAndSchema = durableStateTableCfg.schemaName.fold(profile.quoteIdentifier(durableStateTableCfg.tableName))(
+    schema => s"${profile.quoteIdentifier(schema)}.${profile.quoteIdentifier(durableStateTableCfg.tableName)}")
+
+  private lazy val persistenceIdColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.persistenceId)
+  private lazy val globalOffsetColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.globalOffset)
+  private lazy val revisionColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.revision)
+  private lazy val statePayloadColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.statePayload)
+  private lazy val stateSerIdColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.stateSerId)
+  private lazy val stateSerManifestColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.stateSerManifest)
+  private lazy val tagColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.tag)
+  private lazy val stateTimestampColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.stateTimestamp)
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
```diff
@@ -23,8 +23,22 @@ import akka.persistence.jdbc.config.DurableStateTableConfiguration
...
+      s"${profile.quoteIdentifier(schema)}.${profile.quoteIdentifier(durableStateTableCfg.tableName)}")
+
+  private lazy val persistenceIdColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.persistenceId)
+  private lazy val globalOffsetColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.globalOffset)
+  private lazy val revisionColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.revision)
+  private lazy val statePayloadColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.statePayload)
+  private lazy val stateSerIdColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.stateSerId)
+  private lazy val stateSerManifestColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.stateSerManifest)
+  private lazy val tagColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.tag)
+  private lazy val stateTimestampColumn = profile.quoteIdentifier(durableStateTableCfg.columnNames.stateTimestamp)
```
johanandren: Quoting these seems fine to not do lazily but always on construction as regular fields.

core/src/test/resources/h2-default-mode-application.conf
```diff
@@ -0,0 +1,50 @@
+# Copyright (C) 2019 - 2025 Lightbend Inc. <https://akka.io>
+
+// general.conf is included only for shared settings used for the akka-persistence-jdbc tests
+include "general.conf"
+
+# Same as h2-application.conf but without DATABASE_TO_UPPER=false. This exercises H2 in its
+# default mode, where unquoted identifiers are uppercased — the failure mode for the durable
+# state identifier-quoting regression.
```
johanandren: Was it really a regression, I would expect it to simply never worked.

johanandren (commented):
Looks good with some tiny things

core/src/main/scala/akka/persistence/jdbc/state/DurableStateQueries.scala
JustinPihony: Done in 3378020 — dropped lazy from all the quoting vals.

core/src/test/resources/h2-default-mode-application.conf
JustinPihony: Good catch — reworded the comment in 3378020. The bug has existed since the durable state plugin was added; the existing tests just hid it via DATABASE_TO_UPPER=false.

johanandren (approved):

